PR #1, opened 2020-03-26
using 2d list store layer's _tiles,useful for get tile by coordinate x,y

bung87:
when this get merged , flame also need update `TiledComponent`.

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: tiled
-version: 0.2.1
+version: 0.2.2
 authors:
 - Fernando Ultremare <feroult@gmail.com>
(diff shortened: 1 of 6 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: tiled
-version: 0.2.1
+version: 0.2.2
```
luanpotter, line 2: since this is a breaking change, we should do a bigger release (like 0.3). please revert this change as we can bump the version when we are ready to release all the changes

luanpotter: but please add your change to the CHANGELOG.md file under `<next>`

bung87: ok, done this.

PR #4, opened 2020-04-19
Added support for tile margin and spacing

diegomgarcia:


lib/src/tile.dart
```diff
@@ -29,8 +31,10 @@ class Tile {
   }
 
   Tile(this.tileId, this.tileset) {
-    width = tileset.width;
+    width = tileset.width ;
```
luanpotter, line 34: please remove this space

lib/src/tile.dart
```diff
@@ -10,6 +10,8 @@ class Tile {
 
   int width;
   int height;
+  int spacing;
+  int margin;
 
   Map<String, String> properties = new Map();
```
luanpotter, line 16:
```suggestion
  Map<String, String> properties = {};
```

diegomgarcia: Hi that one wasn't from my commit.

lib/src/tile.dart
diegomgarcia: ops didn't saw that, thanks.

PR #5, opened 2020-04-23
Added support for rotation

Arne1303:
Each tile now remembers in which rotation its meant to be laid out, can be used later on to draw the tile in the correct orientation.
I will also issue a pr for Flame which implements this feature in the "tiled_component" class so that every Tile is rotated Correctly.

### DIFF
--- a/lib/src/layer.dart
+++ b/lib/src/layer.dart
@@ -12,7 +12,7 @@ class Layer {
 
   TileMap map;
   List<List<int>> tileMatrix;
-  List<List<int>> tileRotation;
+  List<List<List<bool>>> tileRotation;
 
   List<List<Tile>> _tiles;
   List<List<Tile>> get tiles {
@@ -55,12 +55,12 @@ class Layer {
   // TMX data format documented here: https://github.com/bjorn/tiled/wiki/TMX-Map-Format#data
   assembleTileMatrix(var bytes) {
     tileMatrix = new List<List<int>>(height);
-    tileRotation = new List<List<int>>(height);
+    tileRotation = new List<List<List<bool>>>(height);
 
     var tileIndex = 0;
     for (var y = 0; y < height; ++y) {
       tileMatrix[y] = new List<int>(width);
-      tileRotation[y] = new List<int>(width);
+      tileRotation[y] = new List<List<bool>>(width);
       for (var x = 0; x < width; ++x) {
         var globalTileId = bytes[tileIndex] |
             bytes[tileIndex + 1] << 8 |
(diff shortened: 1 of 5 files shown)


lib/src/layer.dart
```diff
@@ -67,9 +70,54 @@ class Layer {
...
+        bool flippedVertically = (globalTileId & FLIPPED_VERTICALLY_FLAG) == FLIPPED_VERTICALLY_FLAG;
+        bool flippedDiagonally = (globalTileId & FLIPPED_DIAGONALLY_FLAG) == FLIPPED_DIAGONALLY_FLAG;
+
+        //Translating the flags into rotation
+        if(
+          !flippedHorizontally && 
+          flippedVertically && 
+          flippedDiagonally){
+            tileRotation[y][x] = 1; //90 °
+        }else if(
```
luanpotter, line 83: this is not formatted properly, please do:
```
    if (...) {
```

luanpotter: we can have a method that takes in `globalTileId` and returns the rotation int (1, 2, etc). and here you just do
```
tileRotation[y][x] = foo(globalTileId)
```

lib/src/layer.dart
```diff
@@ -67,9 +70,54 @@ class Layer {
...
+          !flippedHorizontally && 
+          !flippedVertically && 
+          flippedDiagonally){
+            tileRotation[y][x] = 7; //90° + Horizontal flip
+        }else{
+          tileRotation[y][x] = 0; //No rotation
+          /*
+            !flippedHorizontally && 
+            !flippedVertically && 
+            !flippedDiagonally
```
luanpotter, line 118: please no commented out code

lib/src/layer.dart
```diff
@@ -97,12 +146,14 @@ class Layer {
...
       
       cols.asMap().forEach((j, Tile t) {
-        tileId = tileMatrix[j][i];
+        tileId = tileMatrix[i][j];
+        rotation = tileRotation[i][j];
         tile = map.getTileByGID(tileId)
-        ..x = i
-        ..y = j
+        ..x = j
+        ..y = i
```
luanpotter, line 153: why are you changing i/j x/y on this PR? is this related?

lib/src/tile.dart
```diff
@@ -13,6 +13,19 @@ class Tile {
   int spacing;
   int margin;
 
+  /*
+      Tile Rotation (Clockwise)
```
luanpotter, line 17: just curious, did you define this or this a tiled convention?

Arne1303: I did define this, though it was easier to work with a single int value. I could change it, then we would translate it in Tiled Component

lib/src/layer.dart
Arne1303: no sry, in my testing the map was rotated by 90° 
I think it just was a flaw in my testing I'll revert that

lib/src/tile.dart
luanpotter: Fine by me to keep it, I was just curious on the origin and also if it was based off something else I would ask to add a reference on the comment.

lib/src/layer.dart
```diff
@@ -67,9 +70,54 @@ class Layer {
...
         // Read out the flags
-        var flipped_horizontally = (globalTileId & FLIPPED_HORIZONTALLY_FLAG);
-        var flipped_vertically = (globalTileId & FLIPPED_VERTICALLY_FLAG);
-        var flipped_diagonally = (globalTileId & FLIPPED_DIAGONALLY_FLAG);
+        bool flippedHorizontally = (globalTileId & FLIPPED_HORIZONTALLY_FLAG) == FLIPPED_HORIZONTALLY_FLAG;
+        bool flippedVertically = (globalTileId & FLIPPED_VERTICALLY_FLAG) == FLIPPED_VERTICALLY_FLAG;
+        bool flippedDiagonally = (globalTileId & FLIPPED_DIAGONALLY_FLAG) == FLIPPED_DIAGONALLY_FLAG;
+
+        //Translating the flags into rotation
+        if(
```
erickzanardo, line 78: @Arne1303 after a lot of discussion we decided that this code should not be on tiled.dart since this lib is supposed to be kind of a mirror of how the tiled works.

So what tiled.dart must do is to expose these 3 booleans: `flippedHorizontally`, `flippedVertically` and `flippedDiagonally`, instead of the rotation. And the renderer (on our case Flame) should do this logic.

Let us know if you are still interested in continue with this PR, otherwise we can finish it :)

Arne1303: I'm still interested in it :>

lib/src/layer.dart
```diff
@@ -67,12 +70,18 @@ class Layer {
...
+        bool flippedVertically = (globalTileId & FLIPPED_VERTICALLY_FLAG) == FLIPPED_VERTICALLY_FLAG;
+        bool flippedDiagonally = (globalTileId & FLIPPED_DIAGONALLY_FLAG) == FLIPPED_DIAGONALLY_FLAG;
 
-        // Clear the flags
+        // Save rotation flags
+        tileRotation[y][x] = [
+          flippedHorizontally,
+          flippedVertically,
+          flippedDiagonally
+        ];
```
luanpotter, line 82: Let's create a class `Flips` that contains three distinct boolean fields? `horizontal`, `vertical` and `diagonal`? It's hard to know what the elements of the array mean when looking at that on other parts of the code.
Then we can have a List<List<Flips>> on the Layer

luanpotter:
Closing in favour of #9 

Don't worry @Arne1303 , your commits are all there

PR #15, opened 2020-06-25
Bump version and fix warnings

luanpotter:
This

 * bumps version to 0.5.0 for release
 * fixes all warnings currently on pub

### DIFF
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-if [[ $(flutter format --line-length=120 -n .) ]]; then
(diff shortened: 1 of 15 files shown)


scripts/lint.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-if [[ $(flutter format --line-length=120 -n .) ]]; then
+if [[ $(flutter format -n .) ]]; then
```
luanpotter, line 3: sadly pub doesn't respect our 120 preference. I personally find 80 waaay too short but there is an ongoing war on flutter

PR #16, opened 2020-07-01
Fixes bug with int points instead of double for objects

wolfenrain:
There is a bug in the current implementation of the `TmxObject`. 

The `x`, `y`, `width`, `height`, `rotation` and `points` properties of an object in Tiled are not integer based but double based. So if you place an object not exactly on an integer the parser wont be able to parse the values properly and it throws an exception.

By switching to the double parser instead of the integer parser for these properties we can load the tiled map again.

So not only does this PR solve a bug, it also allows us to have more fine tuned positioning of the objects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## [next]
 
-## 0.5.0+1
+## 0.5.1
 * Fixes bug with int points instead of double for objects.
 
 ## 0.5.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## [next]
 
+## 0.5.0+1
+* Fixes bug with int points instead of double for objects.
+
 ## 0.5.0
 
 * Fixes bug with inconsistent indexes crashes for rectangular maps.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## [next]
 
+## 0.5.0+1
```
erickzanardo, line 3: I prefer `0.5.1`

wolfenrain: :-) will do.

lib/src/tmx_object.dart
```diff
@@ -4,11 +4,11 @@ class TmxObject {
   String name;
   String type;
 
-  int x;
-  int y;
-  int width = 0;
-  int height = 0;
-  int rotation = 0;
+  double x;
```
wolfenrain, line 7: I changed the type to `double`. But this introduces a breaking change I suppose?

luanpotter: it does, let's release it under 0.6.0. I can update flame shortly after, I think we are doing 0.24 soon-ish :)

PR #18, opened 2021-01-23
json support

synchronisator:
This PR enables json support for tiled maps.
The compare-test shows, that the test.tmx and the same map as test.json have the same output.
A user jan load a json, but decide to use the tmx strusture instead by converting it.


lib/src/object_group.dart
```diff
@@ -12,6 +12,14 @@ class ObjectGroup {
   Map<String, dynamic> properties = {};
   List<TmxObject> tmxObjects = [];
 
+  ObjectGroup();
+
+
```
luanpotter, line 17: double empty line

lib/src/tile_map_parser.dart
```diff
@@ -107,7 +107,7 @@ class TileMapParser {
       case 'base64':
         return _decodeBase64;
       default:
-        throw 'Incompatible encoding found: $encodingType';
+        throw 'Incompatible encoding found: $encodingType'; //TODO Should support play text 'csv'
```
luanpotter, line 110: plain text?

json_structure.puml
```diff
@@ -0,0 +1,257 @@
+@startuml
```
luanpotter, line 1: what is this language? can we have some background? is this used?

synchronisator: this is only a visualisation. you can use it in your IDE or for example at plantuml.com.
The json_structure is like this:
http://www.plantuml.com/plantuml/svg/fLV1ZkCs3BtdAuoSMXJR3mYAmGABja2nYosm2_IiM8p3ZYmP53tZTD1_B-Guh_AC9qASKqnQUPJ8noVwOrGZshTk4qahRrgAdQdf6tL6Z0QfThlDfdOcncghxCZkTkFFpzQ8xdPVJFT737vRlMogwgTlAkoRQc_gnqP2xsqTN93qXhrMTMWxeHWvU4TFv9B_LyiQ998g-oOXLPHxukM_H-BcgB1eY6p9aM_qc9px45p5_i2UbPBZCqVzvRCvaIJGXygbHw1v6dJqiCrTOVytrSesEqeG9gz6ENZ0_YcX8z7J6RaxMqmn-vkGjoH1B4dcZMgQXiGC7DzwsLiQClUOvIUIbAvfawCFcJXlvZixYgJdlUZPYFdwPxPwp75F7INJzstr2XvBlvYkqf0mV_Xny-zcSxNouSE7HMgyAUrs7GwgkFB_ADTwY2p9MtH3-SZeYq_7tZ-ExAxJK-H_Q2J_q0NHqK_ZOpdtot6kR1sXj_SYORV8GI7WJ1EiKRFSp61TCeRBmsaf5uXwPNFpi8SWhT6S_aOQqfl0mS24ZSR3EH-cecP1CgN3UlRwSCa2aYBc-RL8aA-3PTyi9sicJjoQvghQOMNITIpyEihSUMKOdyUbDlHxHrKu72BfyCPpoZsTgLbF9KM43ertp32pJZEfyugRuk8JusDdXF1gfc0G_IUEl7UqWao5IfSRQYK2bgn7G0NNGynL4Uo_7gnwGTx9EUuYyjwmpVYnp5am4I1TOA__ioXP4uM2EpN1l_VAiQSY-RBzIC0L3X9HspcZ8ynt6iugh3HeaMX01TScE--PCkFcsRrQAvn-kne2IReB9IMgS05Wjw1B6ZI6wU5MMKV0Z7dxu792GHsW-5uFfcMNAypHE6uy85YDunhc8udF5BVZWLowex4Wgl0ZXNv63XeK2tflISwin8AdUSndCTrjXqVDa81v4yruEW_SqosXCE3SCCgxllNGzmDBr4a2VXUsOpCridnPiTVHm2IRkMRotHffsBzxxEdk6hFSajTRvlCOUgcnc_J5a-gFcPf43FjffZuRivcw3htNcsRrbE2lXyEKu_5bIEPyaC-c-6iZVFbkpATxHFZB-ERRvLFXsVWcaiQoJkKCAUZ3fVRRzH1Peku30WlkWm4NxZxLdH2eNa5-KOFgPJv2Bd_L5hxz4AbOAiHRFyJ_9gQbLrYs5wElsehjRRYeuRAEbcK3aHVY5WEM25giA2gFxQETFwNsdpv2WdYIKG_YUsl8DeGLToK1w2EpeIYGhdAISAmYwt60xMeS9FjE69ouN3Fi6rLR6bYGVqoEdOIzsRDZFHMOLmYFVMQpapquxc0TBnQ2ODYzlC6dwrZUMnilsmB0qEGFbV7SpfhxZe41dppJM926bkxzj9R8ZeDZ07XroTvoRzxrELW6G4RN0ioRuoDvsxVkFm00

lib/src/object_group.dart
synchronisator: done

lib/src/tile_map_parser.dart
synchronisator: done

json_structure.puml
wolfenrain: I dont think we need this in the repo. Most IDE's already allow you to generate an overview of your class structure. And otherwise you can easily generate it with any tools.

Now we also have to keep this updated whenever we update our classes. While our classes already define a structure. So sounds to me as double work

lib/src/layer.dart
```diff
@@ -26,6 +26,12 @@ class Layer {
 
   Layer(this.name, this.width, this.height);
 
+
+  @override
+  String toString() {
```
wolfenrain, line 31: I am personally not a fan of using `toString` methods for debugging reasons. If I want to know whats inside an instance I'll just add breakpoint in my debugger.

And now we also have to keep this uptodate. So my personal preference would be to remove them.

What do you think @luanpotter ?

wolfenrain (commented):
Did a quick review, few comments.

lib/src/object_group.dart
```diff
@@ -12,6 +12,14 @@ class ObjectGroup {
   Map<String, dynamic> properties = {};
   List<TmxObject> tmxObjects = [];
 
+  ObjectGroup();
```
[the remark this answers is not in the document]
synchronisator: Removing it leads to a compile error.
The named constructor fromXML removes the default constructor. So i need it here.... or am i wrong?

lib/src/tmx_object.dart
```diff
@@ -20,6 +20,14 @@ class TmxObject {
   List<Point> points = [];
   Map<String, dynamic> properties = {};
 
+  TmxObject();
```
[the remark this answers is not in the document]
synchronisator: same as in ObjectGroup

json_structure.puml
synchronisator: ok, i remove them

lib/src/layer.dart
synchronisator: I use this toString as an easy method to check the results in parser_compare_test.dart
But maybe it would be better to create equals methods... but they had to be changed also, every time the class is changed. What do you think is better?
Or should i make a "big test" and compare every attribute?

lib/src/tile_map_json_parser.dart
```diff
@@ -0,0 +1,1279 @@
...
+  MapJson parse(String string) {
+    // TODO Check Filetype?
+    // TODO Custom parser?
+    // TODO ?.toDouble()
+    final MapJson mapJson = MapJson.fromJson(jsonDecode(string));
+    return mapJson;
+  }
+}
+
+class MapJson {
```
synchronisator, line 15: I have done all the json things in one file.
Do you want it in a directory and every class in separate files?

lib/src/layer.dart
wolfenrain: If the default equal methods are not sufficient for comparisons I would recommend overwriting the equals methods.

lib/src/object_group.dart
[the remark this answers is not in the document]
wolfenrain: What kind of compile error, because I just did a quick test in dartpad and a named constructor is allowed without a default constructor.

Possibly that you are trying to construct these classes using the default constructor, but because they dont exist it gives a compile error?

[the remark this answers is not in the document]
synchronisator: Yes. 
i do:
    final ObjectGroup objectGroup = ObjectGroup();
The error is, that ObjectGroup does not provide a default constructor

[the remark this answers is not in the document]
wolfenrain: You should not do it like that then. If you need a specific usecase for it, you should turn that usecase into a named constructor.

luanpotter:
Sorry for the long delay @synchronisator I had some harsh weeks at work. But lets get back to this as I think json support is very useful. I will re-open and address any remaining comments myself.

synchronisator:
> Sorry for the long delay @synchronisator I had some harsh weeks at work. But lets get back to this as I think json support is very useful. I will re-open and address any remaining comments myself.

I have rewritten everything, but have not yet made a new pr. Look here: https://github.com/synchronisator/tiled.dart/tree/jsonsupport

luanpotter:
@synchronisator ah sounds good! I will wait for your PR then. lmk if I can help with anything :)

lib/src/object_group.dart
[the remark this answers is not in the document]
luanpotter: I think the default constructor is ok if it's used, we have classes on flame with named + default constructors where it makes sense.

lib/src/layer.dart
luanpotter: cant you just assert the properties?
```
   assertEquals(layer.width, 100)
```
etc

synchronisator:
@luanpotter ok, i am not sure about this... its my first github pullrequest ever.
Are my changes, i made in the time after i closed this pr, now in it because you reopened it?
if so, here are some explanations:

The current version is based on a tiled format <= 0.14 (6-7years ago), so it is not working properly with newer tmx files.

The Structure has changed a bit since then and it is (mainly) the same between tmx and json.
The code (tmx and json) is build to match the documentation:
https://doc.mapeditor.org/en/stable/reference/tmx-map-format/

But i made some changes:
TiledMap has some convenience methods
Layer has tileIDMatrix and tileFlips as convenience method (which is a little problematic on infinite maps, because in this case the data is split up in "chunks" which should be used instead of the matrix -> lib/src/layer.dart:204)

A change to the solution before is that  all layers are in one list.
objectgroup and imagelayer are not in separate lists, because i want to keep the layer-order.

All "circle" references are gone (like tileset knows the map)
Example files are fresh opened and saved in Tiled 1.2.1 because some basic things were missing, like tilecount in tilesets.

Because of some changes in the structure the flame_tiled is not working with this version out of the box.

The tests and documentation could be better...

I removed "computeDrawRect" from Tile because it needs more information from other classes like Tileset and i belive this kind of calculation should be placed in flame_tiled.

The code is NOT nullsave

lib/src/tile_map_json_parser.dart
synchronisator: It is changed in separate files

lib/src/layer.dart
synchronisator: this is an old version and the toString is deleted

lib/src/object_group.dart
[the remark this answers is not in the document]
synchronisator: old code, so i will resolve this

lib/src/chunksize.dart
```diff
@@ -0,0 +1,16 @@
+part of tiled;
+
+class Chunksize {
```
luanpotter, line 3: can we rename everything to be camelCase or TitleCase? like ChunkSize, EditorSetting, offsetX, etc. I don't think this affects parsing and it's easier to read.

lib/src/frame.dart
```diff
@@ -0,0 +1,16 @@
+part of tiled;
+
+class Frame {
+  int duration;
+  int tileid;
```
luanpotter, line 5: eg tileId

lib/src/layer.dart
```diff
@@ -1,128 +1,233 @@
...
+  List<TiledObject> objects = [];
+  double offsetx;
+  double offsety;
+  double opacity;
+  List<Property> properties = [];
+  int startx;
+  int starty;
+  String tintcolor;
+  String transparentcolor;
+  String type; // tilelayer, objectgroup, imagelayer or group
```
luanpotter, line 25: should this be an enum?

lib/src/layer.dart
```diff
@@ -1,128 +1,233 @@
...
-    if (element == null) {
-      throw 'arg "element" cannot be null';
+  Layer.fromJson(Map<String, dynamic> json) {
+    compression = json['compression'];
+    encoding = json['encoding'];
+    if (json['chunks'] != null) {
+      chunks = <Chunk>[];
+      json['chunks'].forEach((v) {
+        chunks.add(Chunk.fromJson(v, encoding, compression));
+      });
```
luanpotter, line 116: this could be a `json['chunks'].map` ?

lib/src/layer.dart
```diff
@@ -1,128 +1,233 @@
...
-      assembleTileMatrix(inflatedString);
+    data = json['data'] != null ? decodeData(json['data'], encoding, compression) : [];
+    draworder = json['draworder'];
+    height = json['height'];
+    id = json['id'];
+    image = json['image'];
+    if (json['layers'] != null) {
+      layers = <Layer>[];
+      json['layers'].forEach((v) {
+        layers.add(Layer.fromJson(v));
```
luanpotter, line 126: ditto

luanpotter: and below

lib/src/point.dart
```diff
@@ -0,0 +1,18 @@
+part of tiled;
+
+class Point {
+  num x;
```
luanpotter, line 4: num? should this be int or double?

lib/src/chunk.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  Chunk.fromJson(Map<String, dynamic> json, String encoding, String compression) {
+    data = Layer.decodeData(json['data'], encoding, compression);
+    height = json['height'];
+    width = json['width'];
+    x = json['x'];
+    y = json['y'];
+
+    tileIDMatrix = List.generate(height, (_) => List<int>(width));
+    tileFlips = List.generate(height, (_) => List<Flips>(width));
```
luanpotter, line 35: shouldnt this call Layer.generateTiles?

lib/src/grid.dart
```diff
@@ -0,0 +1,19 @@
+part of tiled;
+
+class Grid {
+  int height;
+  String orientation;
```
luanpotter, line 5: should this be an enum?

lib/src/terrain.dart
```diff
@@ -0,0 +1,30 @@
...
+    });
+  }
+
+  Terrain.fromJson(Map<String, dynamic> json) {
+    name = json['name'];
+    if (json['properties'] != null) {
+      properties = <Property>[];
+      json['properties'].forEach((v) {
+        properties.add(Property.fromJson(v));
+      });
```
luanpotter, line 26: map

lib/src/terrain.dart
```diff
@@ -0,0 +1,30 @@
...
+  List<Property> properties = [];
+  int tile;
+
+  Terrain.fromXml(XmlElement xmlElement) {
+    name  = xmlElement.getAttribute('name');
+    tile  = int.tryParse(xmlElement.getAttribute('tile') ?? '');
+    xmlElement.children.whereType<XmlElement>().forEach((XmlElement element) {
+      switch (element.name.local) {
+        case 'properties':
+          element.nodes.whereType<XmlElement>().forEach((element) {properties.add(Property.fromXml(element));});
```
luanpotter, line 14: map

lib/src/tile.dart
```diff
@@ -1,65 +1,78 @@
...
-
-  // Tile global IDs aren't 1-based, but start from "1" (0 being an "null tile").
+  List<Frame> animation = [];
   int gid;
+  TiledImage image;
+  Layer objectgroup;
+  double probability;
+  List<Property> properties = [];
+  List<int> terrain = []; // index of the terrain
+  String type;
```
luanpotter, line 11: should this be an enum?

lib/src/tiledmap.dart
```diff
@@ -0,0 +1,179 @@
...
+  int hexsidelength;
+  bool infinite;
+  List<Layer> layers = [];
+  int nextlayerid;
+  int nextobjectid;
+  String orientation; // orthogonal, isometric, staggered or hexagonal
+  List<Property> properties = [];
+  String renderorder; // right-down (the default), right-up, left-down and left-up.
+  String staggeraxis; // x or y
+  String staggerindex; // odd or even
```
luanpotter, line 18: enums!

lib/src/tiledmap.dart
```diff
@@ -0,0 +1,179 @@
...
+  String type;
+  num version;
+  int width;
+
+  // Convenience Methods
+  Tile getTileByGID(int cleanTileID){
+    final Tileset tileset = getTilesetByTileID(cleanTileID);
+    final tiles = tileset.tiles.where((element) => element.gid == (cleanTileID - tileset.firstgid)).toList();
+    if(tiles.isNotEmpty){
+      return tiles.first;
```
luanpotter, line 32: `firstWhere(xxx, orElse: () => null)?.first`

luanpotter:
> Are my changes, i made in the time after i closed this pr, now in it because you reopened it?

yes PRs are updated automatically if you push to the same branch.

> Because of some changes in the structure the flame_tiled is not working with this version out of the box.

That is ok, we can update that later (as long as we don't make it impossible, ie, remove features)

> I removed "computeDrawRect" from Tile because it needs more information from other classes like Tileset

can we just move it up the data hierarchy? if not ok to put on flame

> The code is NOT nullsave

that is perfectly fine
--- 

I left some initial comments, this change is very big! But I understand the reason.

I think our goals here are:
* update our underlying data classes to the most up-to-date representation
* make sure parseXml and parseJson works
* make sure we have fallbacks so the old TMX files still work

Now for some general comments:
 * we should never have commented out code. it seems you commented a lot of tests because they use default constructors? feel free to add default constructors, you can mark them with @visibleFroTest. or you can call fromJson on the test and create the json that represents what you want. both work. or you can have a non-default constructor that gets all the fields with some default values (like Point(this.x, this.y)) and use that on the test. regardless, we should keep old tests and never have commented out code
 * make sure we keep an example for the old TMX format not throwing errors
 * you added a few images, is that ok in terms of licensing? ie are they all open source?

lib/src/layer.dart
synchronisator: enums are a pain. Or i am too stupid.
Example:
enum FileEncoding { csv, base64 }

Json Parsing via:
    encoding = FileEncoding.values.firstWhere(
        (e) => e.toString() ==json['encoding'],
        orElse: () => null);
fails, because e.toString() is  "FileEncoding.base64" instead of "base64"

Only for this i had to add a new dependency to package:flutter/foundation.dart to use something like this:
describeEnum(e) == json['encoding']

I am not sure, but i dont think this dependency is worth it.
Enums in dart are not my favorite. Like them in java, but not in dart. Or i have not understood how this should work....

lib/src/chunksize.dart
synchronisator: Done

lib/src/frame.dart
synchronisator: Done

lib/src/layer.dart
synchronisator: nice! Thanks. 
Is done

lib/src/layer.dart
synchronisator: Done

lib/src/terrain.dart
synchronisator: Done

lib/src/terrain.dart
synchronisator: done

lib/src/tiledmap.dart
synchronisator: Done

lib/src/layer.dart
synchronisator: ok, i did it with extensions. But for that i had to change the sdk-version from 
  sdk: ">=2.3.0 <3.0.0"
to
  sdk: ">=2.7.0 <3.0.0"

If have a better idea. Just say it.

lib/src/point.dart
synchronisator: Yeah, thats wierd. if you look at the documentation its double in json and int in tmx:
https://doc.mapeditor.org/en/stable/reference/json-map-format/#point
https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#point
So i decided to use num.

lib/src/chunk.dart
synchronisator: Oh wow. Thanks. Done

lib/src/layer.dart
luanpotter: oh I love it, that is very clever! and the sdk constraint is fine, 2.7 is very stable and flame and flutter all have the 2.7 requirement now

lib/src/point.dart
luanpotter: I see. maybe just add a comment explaining then? because people who use it might need to do some casting.

synchronisator: I asked the people from tiled and I misunderstood the meaning of "in pixels" because it is possible to make sub-pixel measurements. So both is double. I will fix this

synchronisator: Done

synchronisator:
Hi,
sorry for the long time not changing anything. Had much work here...
Now i try to answer all of your questions:

"computeDrawRect": "can we just move it up the data hierarchy? if not ok to put on flame_tiled"

We could add this to "tileset" like this:
`
  Rectangle computeDrawRect(Tile tile) {
    if (tile.image != null) {
      return Rectangle(0, 0, tile.image.width, tile.image.height);
    }
    final row = (tile.gid - firstGId) ~/ columns;
    final column = (tile.gid - firstGId) % columns;
    final x = margin + (column * (tileWidth + spacing));
    final y = margin + (row * (tileHeight + spacing));
    return Rectangle(x, y, tileWidth + spacing, tileHeight + spacing);
  }`

Is this a solution you would like?

synchronisator:
"make sure we have fallbacks so the old TMX files still work"

The old test.tmx failes with the current version of my code. This is because the tilecount is missing in the file. But the tilecount is important to create the tileMatrix.
The old Code is buggy at this point. It uses the tileWidth and tileHeight
```
      width = dsl.intOr('tilewidth', width);
      height = dsl.intOr('tileheight', height);
```
But this is wrong.

I could recreate this buggy behavior if the tileCount is not available but i dont think this is a good solution....

synchronisator:
So. we reach the last three questions:
1) uncommented Code:
 i will fix this ... soon.
2) make sure we keep an example for the old TMX format not throwing errors
I added it again as test_old.tmx
3) you added a few images, is that ok in terms of licensing? ie are they all open source?
I have the maps from three other repos on github:
floortileset.png and tilesets_deviant_milkian_1.png are from Workadventu.re
https://github.com/thecodingmachine/workadventure/tree/develop/maps/Lyon

isometric_grass_and_water.png is from Tiled
https://github.com/mapeditor/tiled/tree/55175d1930b98087672a2f83ac9f30ddb8fed2c4/examples

And coins.png is from flutter_tiled
https://github.com/flame-engine/flame_tiled/tree/master/example/assets/images

If the images are ok, the only open task would be to make the tests clean and without uncommented code. I will work on that in the next time.
If you have more comments please let me know.

synchronisator:
Hi, i updated the tests.
There are 2 problems left.
1) The computeDrawRect is moved to TileSet, but return "other" results and i have no clue why.
2) In tmx_object_test.dart there is a test for tiledObject.isRectangle but there is no value inside the files to see if its a rectangle, and i dont know how to compute this from the polygon. 
Maybe you have an idea.
Thanks

Akida31:
@synchronisator after testing with the TiledMapEditor, rectangles have no certain property set and no polygon/ polyline but only width and height. The width and height could be zero, that would mean that the rectangle has the size of one tile (from what i can see).
Probably it would be good to add the getter tiledObject.isRectangle so that not only the tests but also the users could use this.

The question is if isRectangle should check the polygon if it has 4 points creating a rectangle (which could be done be some math)

synchronisator:
> @synchronisator after testing with the TiledMapEditor, rectangles have no certain property set and no polygon/ polyline but only width and height. The width and height could be zero, that would mean that the rectangle has the size of one tile (from what i can see).
> Probably it would be good to add the getter tiledObject.isRectangle so that not only the tests but also the users could use this.
> 
> The question is if isRectangle should check the polygon if it has 4 points creating a rectangle (which could be done be some math)

@Akida31 
Yes, this is exactly the thing i wrote as my second problem. I dont know how to do this check correct. From where do you have the information about "empty width/height means tilewidth/tileheight"? 
Yes, math would do the "4points creating a rectangle"-job. I was not sure if this is the way i should do that. But if the others think the same way, i will calculate if its a rectangle.

Akida31:
I just created a map with different shapes and looked into the Json-Export-file. Probably not the cleanest way. Would be nice if you could verify that my assumptions are true.
It was an addition to your second question and I thought it would help you to hear an opinion/ some information related to the problem. If you want I could write the implementation the way I would solve that

synchronisator:
@Akida31 Thanks for your testing.
I looks like you are right. If i dont "drag a size" to a rectangle in Tiled it has a default width/height but its not tileheight/tilewidth. In my testmap i have 32x32 Tiles. But the created rectangle is 20x20. its strange.

Something similar is happening with an ellipse without dragging. here is the height 20 again without a height in the json. Maybe its a tiled default.

After a little bit of analysing. I think the rule is: If its NOT a polygon or polyline or ellipse or point it IS a rectangle.
This includes "rotated" rectangles without math calculations.

Question to all of you: Is it needed to calculate rectangles from polygons?

spydon:
@synchronisator great work! How is it going, anything you need help with? :)

Once this is in we will start migrating this package to null-safety.

synchronisator:
@spydon thanks
The last thing is (I quote myself)
"The computeDrawRect is moved to TileSet, but return "other" results and i have no clue why."

If someone has any ideas on this, please comment. I have no idea what is going wrong here...

spydon:
> @spydon thanks
> The last thing is (I quote myself)
> "The computeDrawRect is moved to TileSet, but return "other" results and i have no clue why."
> 
> If someone has any ideas on this, please comment. I have no idea what is going wrong here...

Thanks for the quick reply! I'll have a look with the team.
What does it mean with "other" results btw, do you have an example?

synchronisator:
I am not at my pc right now. But I think there was a test for this method that fails now... But I am not 100% sure. Will have a look later

synchronisator:
@spydon 
It is part of the parser_test.dart line 158 following
i uncommented it.

Other things:
There are 4 other TODOs in the code. Maybe someone has things to say to clear them out.

- tile_test.dart: // TODO no default constructor on Tileset
so the test is diffrent than before, but i think that is ok.

- layer.dart: //TODO zstd compression not supported in dart
There is no zstd compression in dart so i think this is ok as well

- template.dart: // TODO is it possible to have an externel tsx here?
I have no idea if this is possible, but i dont think so...

- tile.dart: //TODO correct?
A Question about json parsing and if the default is correct. This could be good to be reviewed by someone.

luanpotter:
Haven't dug deep into what each one is, but my initial 2cents:

```
// TODO no default constructor on Tileset
can't we just add a default constructor? no issues for me if it's needed

//TODO zstd compression not supported in dart
just leave out this feature for now

// TODO is it possible to have an externel tsx here?
just leave out this option for now

//TODO correct?
that I will take a look into the code to see what it's about haha
```

luanpotter:
Since this PR is so useful and so big, I think we can be a bit lenient with missing features and followups, as long as they are well documented here :)

synchronisator:
Hi,
this PR is waiting now for a long time.
What do i have to change? What do you expect happens next?
Who would approve/merge this?
Thanks

test/rectangle_map_test.dart
```diff
@@ -1,44 +1,48 @@
...
+      expect(
+          layer.tileIDMatrix[5], equals([0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0]));
+      expect(
+          layer.tileIDMatrix[6], equals([0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0]));
+      expect(
+          layer.tileIDMatrix[7], equals([0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0]));
+      expect(
+          layer.tileIDMatrix[8], equals([0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0]));
+      expect(
+          layer.tileIDMatrix[9], equals([0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0]));
```
spydon, line 39: Minor comment, these should have a trailing comma since it is split on two lines.

spydon (approved):

test/rectangle_map_test.dart
luanpotter: we can do that as a followup

luanpotter (approved):
let's merge this and address things as needed

PR #20, opened 2021-03-29
Nullsafety 

Schnurber:
Nullsafety with avoidance of nullable variables. All tests are working.

### DIFF
--- a/lib/src/flips.dart
+++ b/lib/src/flips.dart
@@ -1,5 +1,3 @@
-part of tiled;
-
 class Flips {
   final bool horizontally;
   final bool vertically;
@@ -10,9 +8,9 @@ class Flips {
   const Flips.defaults() : this(false, false, false);
 
   Flips copyWith({
-    bool horizontally,
-    bool vertically,
-    bool diagonally,
+    bool? horizontally,
+    bool? vertically,
+    bool? diagonally,
   }) {
     return Flips(
       horizontally ?? this.horizontally,
--- a/lib/src/flips.dart
+++ b/lib/src/flips.dart
@@ -21,6 +21,7 @@ class Flips {
 
   @override
   String toString() {
-    return 'Flips { horizontally: $horizontally, vertically: $vertically, diagonally: $diagonally }';
+    return 'Flips { horizontally: $horizontally, '
+        'vertically: $vertically, diagonally: $diagonally }';
   }
 }
--- a/lib/src/layer.dart
+++ b/lib/src/layer.dart
@@ -60,7 +60,8 @@ class Layer {
     properties = TileMapParser.parsePropertiesFromElement(element);
   }
 
-  // TMX data format documented here: https://github.com/bjorn/tiled/wiki/TMX-Map-Format#data
+  // TMX data format documented here:
+  // https://github.com/bjorn/tiled/wiki/TMX-Map-Format#data
   void assembleTileMatrix(var bytes) {
     tileMatrix = List.generate(height, (_) => List<int>.filled(width, 0));
     tileFlips = List.generate(height,
@@ -106,8 +107,8 @@ class Layer {
     int px = 0;
     int py = 0;
 
-    _tiles =
-        List.generate(height, (_) => List<Tile>.filled(width, Tile(0, Tileset(0))));
+    _tiles = List.generate(
+        height, (_) => List<Tile>.filled(width, Tile(0, Tileset(0))));
     _tiles!.asMap().forEach((j, List<Tile> rows) {
       px = 0;
 
--- a/lib/src/layer.dart
+++ b/lib/src/layer.dart
@@ -23,11 +23,11 @@ class Layer {
   late List<List<Flips>> tileFlips;
 
   List<List<Tile>>? _tiles;
-  List<List<Tile>>? get tiles {
+  List<List<Tile>> get tiles {
     if (_tiles == null) {
       _recalculateTiles();
     }
-    return _tiles;
+    return _tiles!;
   }
 
   Layer(this.name, this.width, this.height) : visible = true;
@@ -40,7 +40,7 @@ class Layer {
       visible = dsl.boolOr('visible', true);
     });
 
-    final XmlNode dataElement = element.children.firstWhere(
+    final dataElement = element.children.firstWhere(
       (node) => node is XmlElement && node.name.local == 'data',
     );
     if (dataElement is XmlElement) {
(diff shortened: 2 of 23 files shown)


pubspec.yaml
```diff
@@ -1,14 +1,19 @@
 name: tiled
-version: 0.6.0
+version: 0.8.0
```
Akida31, line 2: why this bigger increase?

lib/src/flips.dart
```diff
@@ -10,9 +8,9 @@ class Flips {
   const Flips.defaults() : this(false, false, false);
 
   Flips copyWith({
-    bool horizontally,
-    bool vertically,
-    bool diagonally,
+    bool? horizontally,
+    bool? vertically,
+    bool? diagonally,
```
spydon, line 13: Couldn't these be non-nullable and have default values?

lib/src/layer.dart
```diff
@@ -1,53 +1,54 @@
...
   static const int FLIPPED_DIAGONALLY_FLAG = 0x20000000;
 
-  String name;
-  int width;
-  int height;
-  bool visible;
+  String name = "";
+  int width = 0;
+  int height = 0;
+  bool visible = true;
```
spydon, line 17: Why do these need initialization here when they also come in as non-nullable arguments in the constructor?

lib/src/layer.dart
```diff
@@ -1,53 +1,54 @@
...
-
     NodeDSL.on(element, (dsl) {
       name = dsl.strOr('name', name);
       width = dsl.intOr('width', width);
       height = dsl.intOr('height', height);
       visible = dsl.boolOr('visible', true);
     });
 
-    final dataElement = element.children.firstWhere(
+    final XmlNode dataElement = element.children.firstWhere(
```
spydon, line 43: No explicit type notations on local variables.

lib/src/layer.dart
```diff
@@ -56,13 +57,14 @@ class Layer {
...
+    properties = TileMapParser.parsePropertiesFromElement(element);
   }
 
   // TMX data format documented here: https://github.com/bjorn/tiled/wiki/TMX-Map-Format#data
   void assembleTileMatrix(var bytes) {
-    tileMatrix = List.generate(height, (_) => List<int>(width));
-    tileFlips = List.generate(height, (_) => List<Flips>(width));
+    tileMatrix = List.generate(height, (_) => List<int>.filled(width, 0));
+    tileFlips = List.generate(height,
+        (_) => List<Flips>.filled(width, const Flips(false, false, false)));
```
spydon, line 67: Not part of this PR, but these generate functions should be able to fill the lists with the correct values directly instead of having the loops afterwards.

lib/src/layer.dart
```diff
@@ -1,53 +1,54 @@
...
-  List<List<int>> tileMatrix;
-  List<List<Flips>> tileFlips;
+  TileMap? map;
+  late List<List<int>> tileMatrix;
+  late List<List<Flips>> tileFlips;
 
-  List<List<Tile>> _tiles;
-  List<List<Tile>> get tiles {
+  List<List<Tile>>? _tiles;
+  List<List<Tile>>? get tiles {
```
spydon, line 26: This shouldn't need the `?` right?

lib/src/object_group.dart
```diff
@@ -1,32 +1,32 @@
...
+import 'tile_map_parser.dart';
+import 'tile_map.dart';
+import 'tmx_object.dart';
+import 'node_dsl.dart';
 
 class ObjectGroup {
-  String name;
-  String color;
+  String name= "";
+  String color = "";
```
spydon, line 9: What does it mean for these to default to empty strings?

lib/src/object_group.dart
```diff
@@ -1,32 +1,32 @@
...
       color = dsl.strOr('color', color);
       opacity = dsl.doubleOr('opacity', opacity);
       visible = dsl.boolOr('visible', visible);
     });
+    properties = TileMapParser.parsePropertiesFromElement(element);
 
-    properties = TileMapParser._parsePropertiesFromElement(element);
-
-    final objectNodes = element.children
+    final Iterable<XmlElement> objectNodes = element.children
```
spydon, line 29: No types on local variables.

lib/src/tile.dart
```diff
@@ -1,34 +1,38 @@
...
+  int gid = 0;
 
-  int width;
-  int height;
-  int spacing;
-  int margin;
+  int width = 0;
+  int height = 0;
+  int spacing = 0;
+  int margin = 0;
```
spydon, line 18: These should be late right?

spydon (commented):
Only commented 5 files so far, but I think many of the comments apply to patterns on the other files too. You shouldn't add values to fields that are later initialized in the constructor (add late/final late on them where it makes sense) and don't add explicit types on local variables.

pubspec.yaml
Akida31: and probably update the README according to the version

lib/src/layer.dart
Schnurber: There are two more constructors (onne named constructor ), and then code would be double, I think this is very clear to read.

lib/src/layer.dart
Schnurber: I agree!

lib/src/layer.dart
Schnurber: In this case I think it is a complex data structure, an it comes form somwhere and there is no meaningful simple initial value.

lib/src/object_group.dart
Schnurber: I don't know, null is not much better. ;-)

lib/src/object_group.dart
Schnurber: I agree!

lib/src/tile.dart
Schnurber: Maybe, but there is also `Tile.emptyTile()`, so it is easier, I think.

lib/src/flips.dart
Schnurber: I'm not sure, but bool types could then have three values, true, false, and null.

pubspec.yaml
Schnurber: Sorry...

lib/src/layer.dart
spydon: On this getter we know that it is initialized, so you can just return `_tiles!`

Schnurber: Yes, the `?` could be removed for the getter, but what you want to return instead if `_tiles` is `null`?

spydon: It can't be null in the getter, since it does `_recalculateTiles()` if it is null.

Schnurber: ...it should be `_tiles!`

lib/src/tile.dart
spydon: Then the default values for an empty tile should be in the constructor.

luanpotter:
Let me take a look at this, I will try to update it with current master

Issue #24, opened 2021-06-11
parser.dart not imported in v0.7.0


supagu:
in tiled.dart I see:

`import 'src/parser.dart';`

should this be:

`export 'src/parser.dart';`

as I cant return a Parser from my TsxProvider

supagu:
hrmm, seems to not be a working now. closing.

supagu:
hrmm I'm getting a build error still actually. Re-opening:

```
Launching lib/main.dart on macOS in debug mode...
lib/main.dart:1
/usr/local/Caskroom/flutter/2.0.4/flutter/.pub-cache/hosted/pub.dartlang.org/tiled-0.7.0/lib/src/chunk.dart:45:5: Error: Type 'Parser' not found.
    Parser parser,
    ^^^^^^
/usr/local/Caskroom/flutter/2.0.4/flutter/.pub-cache/hosted/pub.dartlang.org/tiled-0.7.0/lib/src/common/enums.dart:20:35: Error: Type 'Parser' not found.

extension MapOrientationParser on Parser {
                                  ^^^^^^
/usr/local/Caskroom/flutter/2.0.4/flutter/.pub-cache/hosted/pub.dartlang.org/tiled-0.7.0/lib/src/common/enums.dart:55:32: Error: Type 'Parser' not found.
```

Gibbz:
Can this fix be merged please?

luanpotter:
Merged! Sorry for the delay, had a small issue with the build system.

PR #25, opened 2021-06-13
Fixed export of parser.dart

supagu:
PR to fix issue #24


supagu:
Can we get this test re-run and then merged as I have made fixes to this

luanpotter (approved):
I think I intended to make the impl private but that file clearly also has the interface and exceptions which can be necessary

spydon (approved):

PR #26, opened 2021-06-14
Tile by gid fix

supagu:
Patch to fix tileByGid which was returning the wrong tile as firstGid is a 1 based index.

This branch also includes change #25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## [next]
 
+## 0.7.1
+* Expose Parser
+
 ## 0.7.0
 * Adds JSON support
 * Adds Null-safety support
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Include the following in your `pubspec.yaml`:
 
 ```yaml
     dependencies:
-      tiled: 0.7.0
+      tiled: 0.7.1
 ```
 
 ## Usage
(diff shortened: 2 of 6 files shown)


supagu:
Can we get this test re-run and then merged as I have made fixes to this

luanpotter (approved):
Wow, quite confusing that `firstGid` is 1 based but `localId` is not!

spydon (approved):

erickzanardo (approved):

luanpotter:
I fixed the build; but this is actually breaking the tests now.

luanpotter:
Seems at least one test is very clear on checking:

```dart
    test('returns an empty Tile if GID is 0', () {
      tile = map.tileByGid(0);
      expect(tile.isEmpty, isTrue);
    });
```

I  definitely do not know or remember the rationale there. Are you sure about this 1 vs 0 discrepancy?

supagu:
According to this post:
https://discourse.mapeditor.org/t/option-to-export-tile-data-with-0-based-indexes/2672/2

We might need to check for 0 before subtracting 1.

luanpotter:
I think they are both 1 based, by reading that thread:

> Indeed, global ID 0 is used for empty cells so the first tile in the first tileset always gets global ID 1.

0 always mean no tile. both for local and global. so I don't see why the conversion is necessary...

luanpotter:
I tested this change with flame_tiled and it is totally correct. Go figure!

luanpotter:
I merged this on the command line + the test fixes

luanpotter:
actually, that was not the solution. after many more tests, I believe this is the real deal (the problem was something else):
https://github.com/flame-engine/tiled.dart/commit/cb08bc937f82a6ed42cb47fa9b526446ea8e9f76

Issue #31, opened 2021-12-18
ObjectGroup on tiles not getting parsed from .tsx file


lfraker:
I'm new to Flame and Tiled, so I may be doing something wrong, but I am trying to annotate collidable tiles in Tiles Collision Editor to simplify the work of creating repeated collidable game objects. When I import the TileMap via 

```dart
final TiledComponent tiledMap = await TiledComponent.load('tilemap.tmx', Vector2.all(50));
```
The objectGroup value is null on all Tiles, even though in the .tsx file there is an objectgroup node populated:

```xml
 <tile id="30" probability="0.5">
  <objectgroup draworder="index" id="6">
   <object id="11" x="3.25359" y="4.52951" width="25.3907" height="20.2233">
    <ellipse/>
   </object>
  </objectgroup>
 </tile>
```

I was looking through the repo, and it seems like the issue could simply be a camelCase instead of lowercase name in the Tile parser. in https://github.com/flame-engine/tiled.dart/blob/cb08bc937f82a6ed42cb47fa9b526446ea8e9f76/lib/src/tileset/tile.dart  on line 59, the parser tries to parse an 'objectGroup' node instead of an 'objectgroup' node:

```dart
objectGroup: parser.getSingleChildOrNullAs('objectGroup', Layer.parse)
```

Parsing objectGroups works correctly for tileMaps since the layer parsing code path uses the lowercase name (https://github.com/flame-engine/tiled.dart/blob/f719702b6be93edb46f972bceb1c431e40bf0f7d/lib/src/layer.dart on line 246):

```dart
xml.getChildrenAs('objectgroup', Layer.parse)
```

Please let me know if the above camelCase is intended, or if my use is not a supported case?
Is there another recommended way to simplify the defining of collision boxes on multiple terrain tiles?

spydon:
It definitely looks like an error on our side (https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tmx-objectgroup). Do you want to put up a PR that fixes it? :)

lfraker:
@spydon sure thing, will submit one shortly. I didn't see a contributing guide - any process I need to adhere to for the PR, or for something this simple can I just submit the change if all tests pass?

spydon:
> @spydon sure thing, will submit one shortly. I didn't see a contributing guide - any process I need to adhere to for the PR, or for something this simple can I just submit the change if all tests pass?

The latter. :)

lfraker:
[...]

PR #32, opened 2021-12-19
tileset parser fix for ObjectGroup

lfraker:
Fix to support parsing ObjectGroups from TileSets, bug raised in issue: #31
Included test coverage for ObjectGroup on TileSet case.


spydon (approved):
Lgtm, thanks for the contribution!

spydon:
If you bump the sdk from 2.12 to 2.15 in the pubspec file I think that it will solve the failing checks.

lib/src/tileset/tile.dart
```diff
@@ -56,7 +56,7 @@ class Tile {
               .toList() ??
           [],
       image: parser.getSingleChildOrNullAs('image', TiledImage.parse),
-      objectGroup: parser.getSingleChildOrNullAs('objectGroup', Layer.parse),
+      objectGroup: parser.getSingleChildOrNullAs('objectgroup', Layer.parse),
```
luanpotter, line 59: note: important to make sure the this works with json as well.
looking at [the docs[(https://doc.mapeditor.org/en/stable/reference/json-map-format/) it does appear so, but we dont have any json example with that field. would be easy to export you fixture map as json and add a test for that too?

luanpotter (approved):
LGTM

lib/src/tileset/tile.dart
lfraker: @luanpotter, I didn't get a chance to update this today, but I can add a test to include json before completing the PR tomorrow. It looks like it should work, and not too much work to include another test.

lfraker: @spydon did you want me to update the version number still, or just complete the PR with check failures, and you'll merge into your PR?

lfraker: @spydon did you want me to update the version number still. Or did you want me to leave the PR with check failures, and you'll merge into your PR with the updated version?

spydon: I updated this branch to use the new cicd structure and now everything passes. We'd be really happy if you did a PR for the other issues that you found too! :)

PR #36, opened 2022-01-29
Fixed parseTmx for CSV-formatted tilemap layers

doodlezucc:
This pull request fixes parsing of tilemaps formatted in .tmx with layer data encoding set to CSV. It's actually kind of surprising this issue wasn't addressed before because AFAIK this is Tiled's default configuration.

When parsing a layer, its data was assumed to be a list. This already works well for JSON tilemaps. I added support to convert the CSV `String` into a `List<int>`.

This commit also includes a test for this encoding which would fail and throw a `TypeError` before.

### DIFF
--- a/lib/src/layer.dart
+++ b/lib/src/layer.dart
@@ -274,7 +274,14 @@ abstract class Layer {
     }
 
     if (encoding == FileEncoding.csv) {
-      return (data as List).cast<int>();
+      if (data is List) {
+        return data.cast<int>();
+      } else {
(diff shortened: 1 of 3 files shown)


spydon (approved):
Good job, looks good to me :)

spydon:
@lfraker do you want to review this small PR too?

lfraker:
Looks good, awesome to be able to use CSV now as well!

spydon:
I'll release this later with the pending PR. :)

PR #40, opened 2022-03-25
Fixed crash on empty terain

natebot13:
Copy of #39 but coming from a new branch instead of main.

### DIFF
--- a/lib/src/tileset/tile.dart
+++ b/lib/src/tileset/tile.dart
@@ -24,7 +24,7 @@ class Tile {
   double probability;
 
   /// List of indexes of the terrain.
-  List<int> terrain;
+  List<int?> terrain;
 
   TiledImage? image;
   Layer? objectGroup;
@@ -52,7 +52,7 @@ class Tile {
       terrain: parser
               .getStringOrNull('terrain')
               ?.split(',')
-              .map(int.parse)
+              .map(int.tryParse)
               .toList() ??
           [],
       image: parser.getSingleChildOrNullAs('image', TiledImage.parse),


lib/src/tileset/tile.dart
```diff
@@ -52,7 +52,7 @@ class Tile {
       terrain: parser
               .getStringOrNull('terrain')
               ?.split(',')
-              .map(int.parse)
+              .map(int.tryParse)
               .toList() ??
```
spydon, line 56:
```suggestion
              ?.map(int.tryParse)
              ?.toList() ??
```
It's due to the `terrain` array being null right? Not that it contains unparsable data.

lib/src/tileset/tile.dart
```diff
@@ -24,7 +24,7 @@ class Tile {
   double probability;
 
   /// List of indexes of the terrain.
-  List<int> terrain;
+  List<int?> terrain;
```
spydon, line 27:
```suggestion
  List<int> terrain;
```

spydon (commented):
I think the solution is to use the `?.` all the way, can you try if this works for you?

lib/src/tileset/tile.dart
natebot13: Well, I had a tileset from an artist, and the terrain looked like this:
<tile id="0" terrain=",,,0"/>
but now that I've opened it and saved it with the latest version of tiled, that terrain array is no longer there, and looks like is replaced by wangsets, so the bug is no longer applicable. However, this did let me load the tilesets before I had updated them with tiled.

spydon:
@natebot13 should this be closed?

natebot13:
Sure, seems like the latest way tiled uses terrain is with wangsets. Older tilesets will crash if they're attempted to be parsed, but they can simply be saved with a newer version of tiled to change to wangsets.

PR #50, opened 2022-07-09
Making `tile.type` compatible with Tiled 1.9's new Unified Custom Typ…

kurtome:
…es ("class")

https://www.mapeditor.org/2022/06/25/tiled-1-9-released.html

New in tiled 1.9:
```xml
  <tile id="98" class="Slope">
   <properties>
    <property name="LeftTop" type="int" value="0"/>
    <property name="RightTop" type="int" value="8"/>
   </properties>
  </tile>
```

Previously:
```xml
  <tile id="98" type="Slope">
   <properties>
    <property name="LeftTop" type="int" value="0"/>
    <property name="RightTop" type="int" value="8"/>
   </properties>
  </tile>
```

### DIFF
--- a/lib/src/tileset/tile.dart
+++ b/lib/src/tileset/tile.dart
@@ -47,7 +47,7 @@ class Tile {
   static Tile parse(Parser parser) {
     return Tile(
       localId: parser.getInt('id'),
-      type: parser.getStringOrNull('type'),
+      type: parser.getStringOrNull('type') ?? parser.getStringOrNull('class'), /// Tiled 1.9 "type" has been moved to "class"
       probability: parser.getDouble('probability', defaults: 0),
       terrain: parser
               .getStringOrNull('terrain')


spydon (approved):

CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
+## 0.8.4
+* Adding support for `tile.type` using the "class" attribute for Tiled 1.9's Unified Custom Types
+
 ## 0.8.3
 * Downgrade meta dependency
 
-## 0.8.2
-* Add support for class, which is replacing type in tiled 1.9
```
spydon, line 5: You should never remove entries from the changelog, if the entry was wrong it should be explained in the new entry.

pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: tiled
-version: 0.8.3
+version: 0.8.4
```
spydon, line 2: This should normally not be bumped in this PR, but since I'll do a release after this is merged it doesn't matter this time.

spydon:
Didn't #48 already solve this? 
@ufrshubham

ufrshubham:
> Didn't #48 already solve this? @ufrshubham

No, #48 addressed this problem only for the tilemap files. This PR seems to fix it for tileset files as well. I didn't know that tilesets also had type property.

lib/src/tileset/tile.dart
```diff
@@ -47,7 +47,10 @@ class Tile {
   static Tile parse(Parser parser) {
     return Tile(
       localId: parser.getInt('id'),
-      type: parser.getStringOrNull('type'),
+
+      /// Tiled 1.9 "type" has been moved to "class"
+      type: parser.getStringOrNull('type') ?? parser.getStringOrNull('class'),
```
ufrshubham, line 52: Just to make it consistent with #48, you can make it check for `class` first, and then `type`. As more and more people will migrate to Tiled 1.9, chances of finding `class` property instead of `type` are more likely.

kurtome: Fixed the ordering.

I think with Tiled 1.9 you can add `class` to nearly anything, which is why they did the rename. We'd probably need to got through and test this for a bunch of other objects, but Layers for one could be useful. Although it raises the question, should we add a `tiledClass` attribute or keep calling them `type` even though they property in the new Tiled UI is "Class", I guess they named it that because layers already have a "type" which means something else.

> In addition, the “Type” property previously available only for objects and tiles is now available for all data types as the new “Class” property. For consistency, this value is written out as “class” also for objects and tiles, but a project-wide compatibility option is provided to make it still write out as “type”

ufrshubham: For `TiledObject` class I kept `type` attribute as it was and just added a getter called `class_` which returns `type`. That way it won't break anyone's existing code and if someone using tiled with newer map files, tries to find class attribute, `class_` will pop-up in auto-complete list.

ufrshubham (approved):

kurtome:
@ufrshubham I added the `class_` getter to `Tile` the same as you did for `Object`. I also added support for `class_` on all 4 of the `Layer` types and added some unit tests.

kurtome:
Also adding `ImageLayer.repeatX` and `ImageLayer.repeatY`, but I promise I'll stop tinkering with this now

PR #52, opened 2022-09-04
Add parser support for Tiled hex colors as ui.Color

kurtome:
Adding direct support for parsing color hex strings as `dart.ui.Color` objects, which is the primary type the interoperates with Dart rendering libraries and most `Canvas` methods use.

Also, should eventually replace [`_parseTiledColor` in `flame_tiled` `RenderableTileMap`](https://github.com/flame-engine/flame/blob/flame_tiled-v1.7.2/packages/flame_tiled/lib/src/renderable_tile_map.dart#L416-L430).

### DIFF
--- a/lib/src/layer.dart
+++ b/lib/src/layer.dart
@@ -77,6 +77,10 @@ abstract class Layer {
   /// any graphics drawn by this layer or any child layers (optional).
   String? tintColor;
 
+  /// Parsed [Color] that is multiplied with
+  /// any graphics drawn by this layer or any child layers (optional).
+  Color? tintColor_;
+
   /// The opacity of the layer as a value from 0 to 1. Defaults to 1.
   double opacity;
 
@@ -100,6 +104,7 @@ abstract class Layer {
     this.startX,
     this.startY,
     this.tintColor,
+    this.tintColor_,
     this.opacity = 1,
     this.visible = true,
     this.properties = const [],
@@ -123,6 +128,7 @@ abstract class Layer {
     final startX = parser.getIntOrNull('startx');
     final startY = parser.getIntOrNull('starty');
     final tintColor = parser.getStringOrNull('tintcolor');
+    final tintColor_ = parser.getColorOrNull('tintcolor');
     final opacity = parser.getDouble('opacity', defaults: 1);
     final visible = parser.getBool('visible', defaults: true);
     final properties = parser.getProperties();
@@ -160,6 +166,7 @@ abstract class Layer {
           startX: startX,
           startY: startY,
           tintColor: tintColor,
+          tintColor_: tintColor_,
           opacity: opacity,
           visible: visible,
           properties: properties,
@@ -176,7 +183,10 @@ abstract class Layer {
           'draworder',
           defaults: DrawOrder.topDown,
         );
-        final color = parser.getString('color', defaults: '#a0a0a4');
+        final color =
+            parser.getString('color', defaults: ObjectGroup.defaultColorHex);
+        final color_ =
+            parser.getColor('color', defaults: ObjectGroup.defaultColor);
         final objects = parser.getChildrenAs('object', TiledObject.parse);
         layer = ObjectGroup(
           id: id,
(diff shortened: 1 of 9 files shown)


spydon:
I didn't get parts of this, what is the point of having two color fields everywhere?

kurtome:
> I didn't get parts of this, what is the point of having two color fields everywhere?

Just trying to avoid breaking clients of this library that may be using the hex string version of the color fields. So, backwards compatibility 🤷 

  Happy to rename the `color_` fields to a better name if you have an idea, but there is prior art for this naming convention like `class_` https://github.com/flame-engine/tiled.dart/pull/50

spydon:
> Just trying to avoid breaking clients of this library that may be using the hex string version of the color fields. So, backwards compatibility 🤷 

Since this lib is still pre 1.0 I think it would be better to just do a breaking change here than introducing this kind of "code smell".

>   Happy to rename the `color_` fields to a better name if you have an idea, but there is prior art for this naming convention like `class_` https://github.com/flame-engine/tiled.dart/pull/50

The difference between that field and this one is that both those fields exist in the specification though?

> And the main rationale: hex strings are just super cumbersome to use as colors in the Dart/Flutter ecosystem, so it makes sense to parse them to `ui.Color` in my opinion.

Agree.

kurtome:
> Since this lib is still pre 1.0 I think it would be better to just do a breaking change here than introducing this kind of "code smell".

Gotcha, I agree it'll be better to just make the breaking change, sounds good!

I almost forgot that `flame_tiled` can pick when it updates to the bumped version and deal with the breaking changes smoothly.

kurtome:
@spydon I pushed some changes that keep both the `String` and the `Color` version of the fields, but with `colorHex` and `color` as the name of the fields for clarity. I think it makes sense to keep exposing the hex string to clients in case they want to integrate with a different API (especially a web client). What do you think?

lib/src/layer.dart
```diff
@@ -75,7 +75,13 @@ abstract class Layer {
 
   /// Hex-formatted tint color (#RRGGBB or #AARRGGBB) that is multiplied with
   /// any graphics drawn by this layer or any child layers (optional).
-  String? tintColor;
+  String? tintColorHex;
```
spydon, line 78: Is this really needed, if the user wants to get the hex code for a color it should be possible to get it from a `Color` object?

kurtome: There is no clean way to go from `ui.Color` to a hex string, you have to do something like:

```dart
final hexString = "0x${color.value.toRadixString(16).padLeft(8, '0')}"
```

I think it makes sense to keep exposing the hex string to clients in case they want to integrate with a different API (especially a web client). Also, it seems in line with the vision of this library to be a thin wrapper over the parsed `.tmx` file, a client may want to access the raw version of the value from the file.

If you think it bloats the API though, I'm fine with removing the hex string properties.

spydon: Alright, let's keep this then, hopefully we won't have more fields that need to be duplicated later. 😅

spydon (approved):

natebot13:
Adding my two cents to this PR: This change added a dependency on flutter's ui library, making this library no longer a dart only lib. With this, I don't think this library can be used anymore in command line dart programs to parse tiled maps if so needed. I feel like the bridge to flutter specific structures should be in flame_tiled perhaps? Perhaps some external extension methods to build Color objects from the string hex values?

spydon:
> Adding my two cents to this PR: This change added a dependency on flutter's ui library, making this library no longer a dart only lib. With this, I don't think this library can be used anymore in command line dart programs to parse tiled maps if so needed. I feel like the bridge to flutter specific structures should be in flame_tiled perhaps? Perhaps some external extension methods to build Color objects from the string hex values?

It doesn't seem like Flutter was added as a dependency in this PR? The Color object here is from dart:ui right? But I can see that Flutter is in the pubspec, and I agree that we should try to remove that.

natebot13:
Ah, I thought dart:ui was from flutter, but maybe I'm mistaken? Either way, I've been experimenting locally with removing the dependency to dart:ui and was able to get my code generator working, which was failing before from the dart:ui import. And I'm glad there's agreement that the flutter dependency should be taken out!

spydon:
Curious, why was the code generator failing due to that one?

PR #55, opened 2022-09-15
feat!: turning List<Property> into CustomProperties

### LINKED ISSUE #54: Make it easier and more efficient to access a property by name
# What could be improved

Properties on tiles and objects are unique, but are returned as `List<Property>`, which makes it difficult to access a property by name and requires iterating through the entire list each time.

So, I'm suggesting changing `tile.properties` and `object.properties` (and anywhere else applicable) to `Map<String, Property>`, for better access.

# Why should this be improved

The Tiled editor doesn't allow multiple properties with the same name, and accessing the properties by name would better match the Tiled editor. In addition, if the properties are keyed by name it will be more efficient.

# Any risks?

This would be a breaking change.

kurtome:
# Description

See #54 

## Checklist

## Breaking Change

This is a breaking change, but I'm intentionally not using `feat!` because we don't want to bump the major version.

### Migration instructions

Use `properties.named<TypeProperty>('prop_name').value` where applicable. Also, `CustomProperties` implements `Iterable<Property>`.

## Related Issues

Fixes #54

### DIFF
--- a/packages/tiled/lib/src/common/property.dart
+++ b/packages/tiled/lib/src/common/property.dart
@@ -119,13 +119,18 @@ class CustomProperties extends Iterable<Property> {
   ///  - color property -> ui.Color
   ///  - file property -> string (path)
   ///  - object property -> int (ID)
-  T get<T>(String name) {
-    return getProp(name).value as T;
+  T getValue<T>(String name) {
+    return getProperty(name).value as T;
   }
 
   /// Get a typed property by its name
-  T getProp<T extends Property<dynamic>>(String name) {
+  T getProperty<T extends Property<dynamic>>(String name) {
     return byName[name]! as T;
+  }
+
+  /// Get a property by its name
+  Property<dynamic> operator [](String name) {
+    return getProperty(name);
   }
 
   @override
--- a/packages/tiled/lib/src/common/property.dart
+++ b/packages/tiled/lib/src/common/property.dart
@@ -23,7 +23,7 @@ class Property<T> {
     required this.value,
   });
 
-  static Property parse(Parser parser) {
+  static Property<Object> parse(Parser parser) {
     final name = parser.getString('name');
     final type = parser.getPropertyType('type', defaults: PropertyType.string);
 
@@ -100,11 +100,11 @@ class Property<T> {
 /// properties.getProp<IntProperty>('foo') ==
 ///     IntProperty(name: 'foo', value: 3);
 /// ```
-class CustomProperties extends Iterable<Property> {
+class CustomProperties extends Iterable<Property<Object>> {
   static const empty = CustomProperties({});
 
   /// The properties, indexed by name
-  final Map<String, Property> byName;
+  final Map<String, Property<Object>> byName;
 
   const CustomProperties(this.byName);
 
@@ -120,16 +120,16 @@ class CustomProperties extends Iterable<Property> {
   ///  - file property -> string (path)
   ///  - object property -> int (ID)
   T? getValue<T>(String name) {
-    return getProperty(name)?.value as T;
+    return getProperty(name)?.value as T?;
   }
 
   /// Get a typed property by its name
-  T? getProperty<T extends Property<dynamic>>(String name) {
-    return byName[name]! as T;
+  T? getProperty<T extends Property<Object>>(String name) {
+    return byName[name] as T?;
   }
 
   /// Get a property by its name
-  Property? operator [](String name) {
+  Property<Object>? operator [](String name) {
     return byName[name];
   }
 
@@ -139,7 +139,7 @@ class CustomProperties extends Iterable<Property> {
   }
 
   @override
-  Iterator<Property> get iterator => byName.values.iterator;
+  Iterator<Property<Object>> get iterator => byName.values.iterator;
 }
 
 /// [value] is the ID of the object
--- a/packages/tiled/lib/src/common/property.dart
+++ b/packages/tiled/lib/src/common/property.dart
@@ -88,15 +88,43 @@ class Property<T> {
   }
 }
 
+/// A wrapper for a Tiled property set
+///
+/// Accessing an int value
+/// ```dart
+/// properties.get<int>('foo') == 3
+/// ```
+///
+/// Accessing an int property:
+/// ```dart
+/// properties.getProp<IntProperty>('foo') ==
+///     IntProperty(name: 'foo', value: 3);
+/// ```
 class CustomProperties extends Iterable<Property> {
   static const empty = CustomProperties({});
 
+  /// The properties, indexed by name
   final Map<String, Property> byName;
 
   const CustomProperties(this.byName);
 
+  /// Get a property value by its name.
+  ///
+  /// [T] must be the type of the value, which depends on the property type.
+  /// The following Tiled properties map to the follow Dart types:
+  ///  - int property -> int
+  ///  - float property -> double
+  ///  - boolean property -> bool
+  ///  - string property -> string
+  ///  - color property -> ui.Color
+  ///  - file property -> string (path)
+  ///  - object property -> int (ID)
+  T get<T>(String name) {
+    return getProp(name).value as T;
+  }
+
   /// Get a typed property by its name
-  T named<T extends Property<dynamic>>(String name) {
+  T getProp<T extends Property<dynamic>>(String name) {
     return byName[name]! as T;
   }
 
--- a/packages/tiled/lib/src/common/property.dart
+++ b/packages/tiled/lib/src/common/property.dart
@@ -33,8 +33,8 @@ class Property {
 }
 
 extension PropertiesParser on Parser {
-  List<Property> getProperties() {
-    return formatSpecificParsing(
+  Map<String, Property> getProperties() {
+    final properties = formatSpecificParsing(
       (json) => json.getChildrenAs('properties', Property.parse),
       (xml) =>
           xml
@@ -42,5 +42,12 @@ extension PropertiesParser on Parser {
               ?.getChildrenAs('property', Property.parse) ??
           [],
     );
+
+    return properties.groupFoldBy((prop) => prop.name, (previous, element) {
+      if (previous != null) {
+        throw ArgumentError("Can't have two properties with the same name.");
+      }
+      return element;
+    });
   }
 }
--- a/packages/tiled/lib/src/common/property.dart
+++ b/packages/tiled/lib/src/common/property.dart
@@ -12,11 +12,10 @@ part of tiled;
 ///   (default string is “”, default number is 0, default boolean is “false”,
 ///   default color is #00000000, default file is “.” (the current file’s
 ///   parent directory))
-class Property {
+class Property<T> {
   String name;
   PropertyType type;
-  // TODO(luan): support other property types
-  String value;
+  T value;
 
   Property({
     required this.name,
@@ -24,16 +23,148 @@ class Property {
     required this.value,
   });
 
-  Property.parse(Parser parser)
-      : this(
-          name: parser.getString('name'),
-          type: parser.getPropertyType('type', defaults: PropertyType.string),
-          value: parser.getString('value'),
+  static Property parse(Parser parser) {
+    final name = parser.getString('name');
+    final type = parser.getPropertyType('type', defaults: PropertyType.string);
+
+    switch (type) {
+      case PropertyType.object:
+        return ObjectProperty(
+          name: name,
+          value: parser.getInt('value', defaults: 0),
         );
+
+      case PropertyType.color:
+        return ColorProperty(
+          name: name,
+          value: parser.getColor('value', defaults: const Color(0x00000000)),
+          hexValue: parser.getString('value', defaults: '#00000000'),
+        );
+
+      case PropertyType.bool:
+        return BoolProperty(
+          name: name,
+          value: parser.getBool('value', defaults: false),
+        );
+
+      case PropertyType.float:
+        return FloatProperty(
+          name: name,
+          value: parser.getDouble('value', defaults: 0),
+        );
+
+      case PropertyType.int:
+        return IntProperty(
+          name: name,
+          value: parser.getInt('value', defaults: 0),
+        );
+
+      case PropertyType.file:
+        return FileProperty(
+          name: name,
+          value: parser.getString('value', defaults: '.'),
+        );
+
+      case PropertyType.string:
+        final value = parser.formatSpecificParsing((json) {
+          return json.getString('value', defaults: '');
+        }, (xml) {
+          final attrString = parser.getStringOrNull('value');
+          if (attrString != null) {
+            return attrString;
+          } else {
+            // In tmx files, multi-line text property values can be stored
+            // inside the <property> node itself instead of in the 'value'
+            // attribute
+            return xml.element.innerText;
+          }
+        });
+
+        return StringProperty(
+          name: name,
+          value: value,
+        );
+    }
+  }
+}
+
+class CustomProperties extends Iterable<Property> {
+  static const empty = CustomProperties({});
+
+  final Map<String, Property> byName;
+
+  const CustomProperties(this.byName);
+
+  /// Get a typed property by its name
+  T named<T extends Property<dynamic>>(String name) {
+    return byName[name]! as T;
+  }
+
+  @override
+  Iterator<Property> get iterator => byName.values.iterator;
+}
+
+/// [value] is the ID of the object
+class ObjectProperty extends Property<int> {
+  ObjectProperty({
+    required super.name,
+    required super.value,
+  }) : super(type: PropertyType.object);
+}
+
+/// [value] is the color
+class ColorProperty extends Property<Color> {
+  final String hexValue;
+
+  ColorProperty({
+    required super.name,
+    required super.value,
+    required this.hexValue,
+  }) : super(type: PropertyType.color);
+}
+
+/// [value] is the string text
+class StringProperty extends Property<String> {
+  StringProperty({
+    required super.name,
+    required super.value,
+  }) : super(type: PropertyType.string);
+}
+
+/// [value] is the path to the file
+class FileProperty extends Property<String> {
+  FileProperty({
+    required super.name,
+    required super.value,
+  }) : super(type: PropertyType.file);
+}
+
+/// [value] is the integer number
+class IntProperty extends Property<int> {
+  IntProperty({
+    required super.name,
+    required super.value,
+  }) : super(type: PropertyType.int);
+}
+
+/// [value] is the double-percision floating-point number
+class FloatProperty extends Property<double> {
+  FloatProperty({
+    required super.name,
+    required super.value,
+  }) : super(type: PropertyType.float);
+}
+
+/// [value] is the boolean
+class BoolProperty extends Property<bool> {
+  BoolProperty({
+    required super.name,
+    required super.value,
+  }) : super(type: PropertyType.bool);
 }
 
 extension PropertiesParser on Parser {
-  Map<String, Property> getProperties() {
+  CustomProperties getProperties() {
     final properties = formatSpecificParsing(
       (json) => json.getChildrenAs('properties', Property.parse),
       (xml) =>
@@ -43,11 +174,13 @@ extension PropertiesParser on Parser {
           [],
     );
 
-    return properties.groupFoldBy((prop) => prop.name, (previous, element) {
-      if (previous != null) {
-        throw ArgumentError("Can't have two properties with the same name.");
-      }
+    // NOTE: two properties should never have the same name, if they do
+    // one will simply override the other
+    final byName =
+        properties.groupFoldBy((prop) => prop.name, (previous, element) {
       return element;
     });
+
+    return CustomProperties(byName);
   }
 }
--- a/packages/tiled/lib/src/common/property.dart
+++ b/packages/tiled/lib/src/common/property.dart
@@ -97,8 +97,13 @@ class Property<T> {
 ///
 /// Accessing an int property:
 /// ```dart
-/// properties.getProp<IntProperty>('foo') ==
+/// properties.getProperty<IntProperty>('foo') ==
 ///     IntProperty(name: 'foo', value: 3);
+/// ```
+///
+/// You can also use an accessor:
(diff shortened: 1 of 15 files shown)


packages/tiled/lib/src/common/property.dart
```diff
@@ -33,14 +33,21 @@ class Property {
...
       (xml) =>
           xml
               .getSingleChildOrNull('properties')
               ?.getChildrenAs('property', Property.parse) ??
           [],
     );
+
+    return properties.groupFoldBy((prop) => prop.name, (previous, element) {
+      if (previous != null) {
+        throw ArgumentError("Can't have two properties with the same name.");
```
jtmcdole, line 48: This seems like a [state error](https://api.flutter.dev/flutter/dart-core/StateError-class.html)

jtmcdole (commented):
👍🏻

ufrshubham:
Title should use `feat!` as it is a breaking change

packages/tiled/lib/src/common/property.dart
ufrshubham, line 48: I don't think we should throw an error here. If `Tiled` does not even allow creation of multiple properties with same name, this case should ideally never occur. Our input condition is `a valid tmx file`.

In the rare case someone manually tweaks their tmx and adds multiple properties with same name, instead of failing, we should follow what Tiled does. Tiled load the map and displays only one of the duplicate properties (the latest that it finds). We can also do that by replacing the value with latest value with a duplicate key is found.

If we really want to go the extra mile and inform the users, it can be done using an assert.

packages/tiled/lib/src/common/property.dart
```diff
@@ -33,14 +33,21 @@ class Property {
 }
 
 extension PropertiesParser on Parser {
-  List<Property> getProperties() {
-    return formatSpecificParsing(
+  Map<String, Property> getProperties() {
```
ufrshubham, line 36: Can we minimize the breaking change by adding a new `getPropertiesMap()` that returns `Map<String, Property>` and making `getProperties()` return the `values` iterable  from the map?

ufrshubham (commented):
Just a few thoughts to make the changes a little soft for the end users

packages/tiled/lib/src/common/property.dart
kurtome: I don't think anyone uses this `getProperties` method directly, but rather the `properties` public variable in `Tile` `Object` `Layer` etc.

We definitely could change the API to:
```dart
class Tile {
  ...

  Iterable<Property> get properties => propertiesByName.values;
  final Map<String, Property> propertiesByName;
}
```

But personally I think that's a worse API and there is really no value in an `Iterable<Property>` accessor, especially when you can get it via `.values`. 

I was under the impression that we're in early stages with this library and want to do large refactors to polish the API before 1.0 so that's why I was leaning towards the more aggressive refactor. But, I'm happy either way 🤷

kurtome:
> Title should use `feat!` as it is a breaking change

@ufrshubham Doesn't that bump this library to 1.0? I didn't want to keep bumping the major version every time we refactor something now.

packages/tiled/lib/src/common/property.dart
ufrshubham: > I don't think anyone uses this `getProperties` method directly, but rather the `properties` public variable in `Tile` `Object` `Layer` etc.

I get what you mean, but this is still just an assumption. If it is a public API, it is always better to assume that it is used by someone. Also, even if `getProperties` is not used by anyone, we are still breaking `properties`.

> But personally I think that's a worse API and there is really no value in an `Iterable<Property>` accessor, especially when you can get it via `.values`.
> 
> I was under the impression that we're in early stages with this library and want to do large refactors to polish the API before 1.0 so that's why I was leaning towards the more aggressive refactor. But, I'm happy either way 🤷

@spydon can comment about plans for 1.0, but I feel making a breaking change and trying not to bump up the major version are conflicting goals here. If you make the changes non-breaking, it will automatically solve the versioning problem. A middle ground would be to let `getProperties` and `properties` work as they used to with a `@deprecated` tag saying that it will be removed in 1.0.

*I might have a hidden motive for making it non-breaking, because `properties` is used in my simple platformer series. It will be better if maintainers should what to do about this 😅

spydon: It doesn't need to have a major bump when it is breaking and sub v1. I think that we can deprecate `getProperties`, but not remove it until we go to past v1, `properties` is more inline with the dart naming convention anyways.
So then your tutorials wouldn't break for a long time.

Poke me when you feel that a new tiled + flame_version should be released btw.

kurtome: So, keep `getProperties` and `properties` and add `getPropertiesByName` and `propertiesByName`?

To be clear `getProperties` and `properties` are two different things, `getProperties` is the function that parses the data which eventually gets set on the `properties` class level vars.

spydon: Aah right, hard to review PRs properly on the phone. Should `getProperties` only be used internally, or is there a value in exposing that to the user? If it's not I would name it `parseProperties` and mark it as `@internal`.
And for `properties` I think that we should do a breaking change on the type then and not pollute the code with unnecessary members already.

kurtome: Sounds good, do you want me to rename ALL the parser methods? 

- Parser.getString
- Parser.getStringOrNull
- Parser.getInt
- Parser.getColor
- etc etc

kurtome: Not sure, I can take a look.

For the method naming, I think the classname makes it self explanatory so I'm inclined to leave it as is, what do you think?

spydon: > For the method naming, I think the classname makes it self explanatory so I'm inclined to leave it as is, what do you think?

Agree, missed that it was on the parser class when I reviewed on mobile.

spydon:
> > Title should use `feat!` as it is a breaking change
> 
> @ufrshubham Doesn't that bump this library to 1.0? I didn't want to keep bumping the major version every time we refactor something now.

It won't, sub v1 breaking changes doesn't lead to v1. One is meant to have breaking changes before v1. So it is good to have the ! for clarity when people are reading the changelog since it will organise the breaking changes in its own section.

kurtome:
@spydon @ufrshubham @jtmcdole Not sure about this API, but it has some merit:

```dart
bool isFoo = tile.properties.named<BoolProperty>("foo").value;
```

The benefit here is each `Property` class can add things, that alternative of making it just `Property<T>` seemed messier.

jtmcdole:
I guess the question is, does the user care about the property or care about the value ? I.e. `<bool>` vs boolproperty...value.

jtmcdole:
Is properties not being turned into a map?

kurtome:
> I guess the question is, does the user care about the property or care about the value ? I.e. `<bool>` vs boolproperty...value.

They definitely care more about the value, the question is how to handle complex properties like File and Object and perhaps others with the same API as primitives

They are being read into a Map, but with a wrapper class so we can add convenience methods to it. We could make it indexable though

ufrshubham:
Value for File and Object properties are just `string`  and `int` respectively. So it is just the custom property types that we need to care about

kurtome:
> Value for File and Object properties are just `string`  and `int` respectively. So it is just the custom property types that we need to care about

We don't want to give them the option of a reference to the Object directly from the property?

kurtome:
If we want to be super minimal we can just give them a map of string/dynamic and make them cast too use the value. Feels gross though

ufrshubham:
> We don't want to give them the option of a reference to the Object directly from the property?

We do, but I'd prefer it to be like this: 
```dart
final isFoo = tile.properties.named<bool>("foo").value;
final objectOfMyClass = tile.properties.named<MyClass>("bar").value;
```
And in the second case, `MyClass` is expected to handle the string/dynamic map. It will nicely hook into the code generation that we were discussing on discord.

kurtome:
I see, I think the other point @jtmcdole is making is, can we just return the value instead of making them call `.value`, we should probably support both

ufrshubham:
> can we just return the value instead of making them call `.value`

Yeah, I am in favor of returning directly the value. We can keep `.value` around if it has some use case that directly returning value does not solve.

ufrshubham:
Here is a rough idea that I had in my mind.

We can have 2 abstract classes for handling/building the custom types created in Tiled
```dart
abstract class CustomPropertyType {}

abstract class CustomPropertyBuilder<T extends CustomPropertyType> {
  T build(Property<dynamic> property);
}
```

Then, in the `CustomProperties` class, we can add a map that stores all the user defined builders.
```dart
class CustomProperties extends Iterable<Property> {
  // omitted code....
  final Map<Type, CustomPropertyBuilder<CustomPropertyType>> builders;

  // Just an example. This method can be moved at a higher level 
  // to make it more accessible to users.
  void register<T extends CustomPropertyType>(
    CustomPropertyBuilder<T> instance,
  ) {
    builders[T.runtimeType] = instance;
  }

  /// Get a typed property by its name
  T named<T>(String name) {
    // If user is requesting a custom property, we look for the registered builder 
    // and delegate the conversion to that builder.
    if (T is CustomPropertyType) {
      final builder = builders[T.runtimeType];
      assert(
        builder != null,
        'No builder found for $T. Did you forgot to register one?',
      );
      return builder!.build(byName[name]!) as T;
    } else {
      return byName[name]! as T;
    }
  }
}
```

If users want to implement their own builders they can do so. Or else they can use code generation to generate derived classes from `CustomPropertyType` and `CustomPropertyTypeBuilder`.

kurtome:
I renamed the methods inside `CustomProperties` so now getting values simpler:

Accessing an int value
```dart
properties.get<int>('foo') == 3
```

Accessing an int property:
```dart
properties.getProp<IntProperty>('foo') == IntProperty(name: 'foo', value: 3);
```

packages/tiled/lib/src/common/property.dart
```diff
@@ -88,15 +88,43 @@ class Property<T> {
...
+  ///  - color property -> ui.Color
+  ///  - file property -> string (path)
+  ///  - object property -> int (ID)
+  T get<T>(String name) {
+    return getProp(name).value as T;
+  }
+
   /// Get a typed property by its name
-  T named<T extends Property<dynamic>>(String name) {
+  T getProp<T extends Property<dynamic>>(String name) {
```
spydon, line 127: We have a ban on abbreviations 😉

kurtome:
> Here is a rough idea that I had in my mind.
> 
> 
> 
> We can have 2 abstract classes for handling/building the custom types created in Tiled
> 
> ```dart
> 
> abstract class CustomPropertyType {}
> 
> 
> 
> abstract class CustomPropertyBuilder<T extends CustomPropertyType> {
> 
>   T build(Property<dynamic> property);
> 
> }
> 
> ```
> 
> 
> 
> Then, in the `CustomProperties` class, we can add a map that stores all the user defined builders.
> 
> ```dart
> 
> class CustomProperties extends Iterable<Property> {
> 
>   // omitted code....
> 
>   final Map<Type, CustomPropertyBuilder<CustomPropertyType>> builders;
> 
> 
> 
>   // Just an example. This method can be moved at a higher level 
> 
>   // to make it more accessible to users.
> 
>   void register<T extends CustomPropertyType>(
> 
>     CustomPropertyBuilder<T> instance,
> 
>   ) {
> 
>     builders[T.runtimeType] = instance;
> 
>   }
> 
> 
> 
>   /// Get a typed property by its name
> 
>   T named<T>(String name) {
> 
>     // If user is requesting a custom property, we look for the registered builder 
> 
>     // and delegate the conversion to that builder.
> 
>     if (T is CustomPropertyType) {
> 
>       final builder = builders[T.runtimeType];
> 
>       assert(
> 
>         builder != null,
> 
>         'No builder found for $T. Did you forgot to register one?',
> 
>       );
> 
>       return builder!.build(byName[name]!) as T;
> 
>     } else {
> 
>       return byName[name]! as T;
> 
>     }
> 
>   }
> 
> }
> 
> ```
> 
> 
> 
> If users want to implement their own builders they can do so. Or else they can use code generation to generate derived classes from `CustomPropertyType` and `CustomPropertyTypeBuilder`.

This is compelling, although I can't help but think we should be using the property's inherent type somehow. Also would be good to not need to rebuild the result every find it's accessed

ufrshubham:
> This is compelling, although I can't help but think we should be using the property's inherent type somehow.

But how do we know what is the inherent type of a custom type that a user has defined in Tiled?


> Also would be good to not need to rebuild the result every find it's accessed

This could be solved by caching the custom property inside `CustomProperties` once it is built.

packages/tiled/lib/src/common/property.dart
jtmcdole: Which is why I was saying "as<T>"....

`CustomProperties` could also just implement:

```dart
Property operator[] (String name) => byName[name];
```

Then you get:
```dart
properties['fu']; // Returns IntProperty!
```
And if you extend `Property` to have `T as<T>()` it looks & feels like other code in Flame.

```dart
properties['fu'].as<int>(); // Returns integer!
```

ufrshubham: In that case, wouldn't it be smart to not have `CustomProperties` class at all. The internal `byName` map will also provide similar API.

----

> ```dart
> properties['fu']; // Returns IntProperty!
> ```

Technically, it will return `Property`. Users will have to manually cast it to `IntProperty`. In that regard, `properties.getProp<IntProperty>('foo')` feels much better.

--- 
We should also try to think about making it easier for end users to find these APIs. We can document the overloaded `[]` operator and `as<>()` as much as we want, but still there will be people who will not read the docs 😅. I feel that explicitly defined methods are easier to find because of code completion in IDEs.

spydon: I think the overloaded operator looks better, I don't think users will have much more of a problem understanding that one than to see that there is the possibility to put generics on `getProp` (which should be named `getProperty` if we decide to go with that API anyways).

Is it it possible to put generics on the operator instead of using `as`?

kurtome: > In that case, wouldn't it be smart to not have `CustomProperties` class at all. The internal `byName` map will also provide similar API.
> 
> 
> 
> ----
> 
> 
> 
> > ```dart
> 
> > properties['fu']; // Returns IntProperty!
> 
> > ```
> 
> 
> 
> Technically, it will return `Property`. Users will have to manually cast it to `IntProperty`. In that regard, `properties.getProp<IntProperty>('foo')` feels much better.


I think the idea is that `as<int>` could be called on the base class so you don't have to cast the property first

> 
> 
> --- 
> 
> We should also try to think about making it easier for end users to find these APIs. We can document the overloaded `[]` operator and `as<>()` as much as we want, but still there will be people who will not read the docs 😅. I feel that explicitly defined methods are easier to find because of code completion in IDEs.

This was my primary reason for not just using Map style interface in the first place, but we could support both. However, `Property<T>.as<int>‘ returning the value itself feels like it's mixing levels of abstraction, it's operating on the value not in the property, is that weird?

jtmcdole: Operator[] is more natural to me. It can't have genetics to my knowledge.

Operator[] retiring the super instance `property` and the user casting to IntProperty (or runtime testing) is also pretty natural. This also grants access to the new custom classes in unparsed form maybe?


as<T> I feel belongs to property and not the map holding the properties.

kurtome: At that point, why not just `properties["foo"].value as int` why have a special method?

kurtome:
@spydon @ufrshubham @jtmcdole I renamed the methods to: `getValue<T>` and `getProperty<T>`.

Also, I added `["name"]` accessor which returns a `Property<dynamic>`.

I think for now we should roll with this, and then subsequent PRs can continue fiddling with the API? I'm not opposed to any of the suggestions above.

spydon:
> Also, I added `["name"]` accessor which returns a `Property<dynamic>`.

Afaik one should pretty much never use `dynamic` nowadays, but instead use `Object?`.

spydon:
> Hm, I tried to update everywhere this was using `dynamic` and I started getting compile errors in other places, for example:
> image
> 
> 
> Also, making the return values nullable and adding `bool has(String name)`

Ah `properties` is a map from string to dynamic? Maybe that could be changed later.

packages/tiled/lib/src/common/property.dart
```diff
@@ -12,35 +12,213 @@ part of tiled;
...
+  ///  - color property -> ui.Color
+  ///  - file property -> string (path)
+  ///  - object property -> int (ID)
+  T? getValue<T>(String name) {
+    return getProperty(name)?.value as T;
+  }
+
+  /// Get a typed property by its name
+  T? getProperty<T extends Property<dynamic>>(String name) {
+    return byName[name]! as T;
```
spydon, line 128: These methods don't need a nullable return type since they are both forced to an object and then cast to T. Or the return should be written in a way that can return null. I wonder why the analyzer didn't pick up on this. 🤔

kurtome: I meant to remove the `!`, they should be nullable in reality, right? It's map of unknown things and the caller may not know if this particular property was set on the tile/obj/etc

spydon: Just removing the `!` is not enough though since you'd cast the value to a non-nullable type so you'll have to change to `as T?` too.

kurtome:
> > > Also, I added `["name"]` accessor which returns a `Property<dynamic>`.
> > 
> > 
> > Afaik one should pretty much never use `dynamic` nowadays, but instead use `Object?`.
> 
> Hm, I tried to update everywhere this was using `dynamic` and I started getting compile errors in other places, for example: image
> 
> Also, making the return values nullable and adding `bool has(String name)`

kurtome:
Oops

packages/tiled/lib/src/common/property.dart
kurtome: 👍  Fixed

kurtome:
> Ah properties is a map from string to dynamic? Maybe that could be changed later.
Fixed, the problem was `Property.parse` was returning `Property` which default the generic to `dynamic`, so now it's returning `Property<Object>`

packages/tiled/lib/src/common/property.dart
```diff
@@ -12,35 +12,213 @@ part of tiled;
...
+/// A wrapper for a Tiled property set
+///
+/// Accessing an int value
+/// ```dart
+/// properties.get<int>('foo') == 3
+/// ```
+///
+/// Accessing an int property:
+/// ```dart
+/// properties.getProp<IntProperty>('foo') ==
```
spydon, line 100:
```suggestion
/// properties.getProperty<IntProperty>('foo') ==
```

packages/tiled/lib/src/common/property.dart
```diff
@@ -12,35 +12,213 @@ part of tiled;
...
+  ///  - string property -> string
+  ///  - color property -> ui.Color
+  ///  - file property -> string (path)
+  ///  - object property -> int (ID)
+  T? getValue<T>(String name) {
+    return getProperty(name)?.value as T?;
+  }
+
+  /// Get a typed property by its name
+  T? getProperty<T extends Property<Object>>(String name) {
```
spydon, line 127:
```suggestion
  T? getProperty<T extends Property<T>>(String name) {
```

packages/tiled/lib/src/common/property.dart
```diff
@@ -12,35 +12,213 @@ part of tiled;
...
+  bool has(String name) {
+    return byName.containsKey(name);
+  }
+
+  @override
+  Iterator<Property<Object>> get iterator => byName.values.iterator;
+}
+
+/// [value] is the ID of the object
+class ObjectProperty extends Property<int> {
```
spydon, line 146:
```suggestion
class ObjectProperty extends Property<Object> {
```

spydon (approved):
Just a few small comments, but other than those lgtm!

packages/tiled/lib/src/common/property.dart
kurtome: `T` is the type of the property here, not the value, so this change would require that the Property has a value of Property recursively. We'd have to add a second generic to the method, `K`, which would be awkward:
`getProperty<IntProperty, int>('foo')`

packages/tiled/lib/src/common/property.dart
kurtome: This is correct actually, the implementation at this stage only give you access to the Tiled object's integer ID

packages/tiled/lib/src/common/property.dart
spydon: Ah, of course. Wouldn't one want to just want to specify the type of the value instead, and then the return type can be `Property<T>?`?

kurtome: Well the intent was to make it simple to get an `IntProperty` or ` ColorProperty`, otherwise you still have to cast it to one of those if you want the actual property type, which would be more relevant if the property has unique methods on it. Otherwise there is really no reason to be grabbing the property itself at all, you can just grab the value directly

spydon: Hmm, what is the real point of having the property wrapper actually? Can't we just offer extensions on the value type instead? Like `toHex` on `Color` for example?

kurtome: We may not need it, but it's to entirely clear to me that every property can just be a simple name/value pair where the value doesn't need custom behavior, if so we should probably ditch the entire Property class

kurtome: For all the `value` based use cases your describing, that's what `getValue<T>` is for, right?

spydon: So, are there any usecases for having this wrapper class? @ufrshubham @jtmcdole

kurtome: I'll also say that there is some value to me to making the class structure match the Tiled data structures for consistency with Tiled itself. Also, we really do want to know if a property is a file string or a text string, etc

spydon (approved):
I think this has been sitting and waiting for more reviews for too long now, I'll merge it.
If anything comes up it can be changed in another PR.

PR #59, opened 2022-10-27
fix: template not parsed

### LINKED ISSUE #47: The Template cannot be properly parsed.
static TiledObject parse(Parser parser) {
    ...
    final template = parser.getSingleChildOrNullAs('template', Template.parse);
    ...
}

  static Template parse(Parser parser) {
    return Template(
      tileSet: parser.getSingleChildOrNullAs('tileset', Tileset.parse),
      object: parser.getSingleChildOrNullAs('object', TiledObject.parse),
    );
  }

Template is provided as a file, but the file is not loaded here.

trunghq3101:
# Description

Currently, `template` is parsed from a child node like this:
```
final template = parser.getSingleChildOrNullAs('template', Template.parse);
```
But according to the current Tiled format, `template` is not a child node anymore:
```
  <object id="31" template="card.tx" name="bank" gid="0" x="-256" y="736" rotation="0"/>
```
There is no place to load the template file as well.
So, it is never found, that's why the `template` field in a parsed `TiledObject` is always `null`. I believe it's related to the issue https://github.com/flame-engine/tiled.dart/issues/47.

The object `id` is also not required anymore:
```
<?xml version="1.0" encoding="UTF-8"?>
<template>
 <object width="288" height="416"/>
</template>
```

## Checklist

## Breaking Change

## Related Issues

Fixes #47 

### DIFF
--- a/packages/tiled/lib/src/objects/tiled_object.dart
+++ b/packages/tiled/lib/src/objects/tiled_object.dart
@@ -44,7 +44,7 @@ part of tiled;
 /// Can contain at most one: <properties>, <ellipse> (since 0.9),
 /// <point> (since 1.1), <polygon>, <polyline>, <text> (since 1.0)
 class TiledObject {
-  int id;
+  int? id;
   String name;
   String type;
   int? gid;
@@ -59,7 +59,7 @@ class TiledObject {
   bool point;
   bool rectangle;
 
-  Template? template;
+  Future<Template?>? template;
   Text? text;
   bool visible;
 
@@ -72,7 +72,7 @@ class TiledObject {
   String get class_ => type;
 
   TiledObject({
-    required this.id,
+    this.id,
     this.name = '',
     this.type = '',
     this.gid,
@@ -105,7 +105,7 @@ class TiledObject {
     final width = parser.getDouble('width', defaults: 0);
     final rotation = parser.getDouble('rotation', defaults: 0);
     final visible = parser.getBool('visible', defaults: true);
-    final id = parser.getInt('id');
+    final id = parser.getIntOrNull('id');
     final gid = parser.getIntOrNull('gid');
     final name = parser.getString('name', defaults: '');
 
@@ -126,7 +126,8 @@ class TiledObject {
       (xml) => xml.getChildren('point').isNotEmpty,
     );
     final text = parser.getSingleChildOrNullAs('text', Text.parse);
-    final template = parser.getSingleChildOrNullAs('template', Template.parse);
+    final template =
+        parser.getExternalPropertyOrNull('template', Template.parse);
     final properties = parser.getProperties();
 
     final polygon = parsePointList(parser, 'polygon');
--- a/packages/tiled/lib/src/tile_map_parser.dart
+++ b/packages/tiled/lib/src/tile_map_parser.dart
@@ -10,12 +10,19 @@ class TileMapParser {
   ///
   /// Accepts an optional list of external TsxProviders for external tilesets
   /// referenced in the map file.
-  static TiledMap parseTmx(String xml, {List<TsxProvider>? tsxList}) {
+  static TiledMap parseTmx(
+    String xml, {
+    List<TsxProvider>? tsxList,
+    Future<TsxProvider> Function(String key)? tsxProviderFunction,
+  }) {
     final xmlElement = XmlDocument.parse(xml).rootElement;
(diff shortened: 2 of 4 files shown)


packages/tiled/lib/src/objects/tiled_object.dart
```diff
@@ -59,7 +59,7 @@ class TiledObject {
   bool point;
   bool rectangle;
 
-  Template? template;
+  Future<Template?>? template;
```
spydon, line 62: Why does this need to be a future?

trunghq3101: The template object must wait for the template file to be loaded.
We also can’t load it beforehand because we can’t know which file to load until reaching a template property.

spydon: But it will be loaded while parsing right? Since this is mutable for some reason it could just be set once it is finished?

trunghq3101: Yes it would be the easier way. However, all the call stack down to TiledObject.parse is synchronous, if we make it async, we must change the upstream too. I want to avoid a breaking change.

spydon: But since it is mutable you don't have to make it async, just set it when it is done?

trunghq3101: Hmm, this new TiledObject instance is added to the final parsed object after `load()` function is completed. If the `template` is mutated after that, users can't know when it happens to get the latest value. If using Future, at least they know explicitly that they need to `await` it to get the value. That's my intention. But I have to admit that it's not the best solution.
I've just thought about another solution. If `load()` function somehow can await for all pending Futures complete, it's the best.

spydon: > I've just thought about another solution. If `load()` function somehow can await for all pending Futures complete, it's the best.

Yeah, that sounds much better, that shouldn't be a problem to do.

kurtome:
>But according to the current Tiled format, template is not a child node anymore

Can you clarify what docs and when this changed? If it changed relatively recently we may want to check for the template xml the old way, and then if it's not there, check the new way. That would support both versions of Tiled. I would also add inline comments that link to the relevant Tiled docs and version changes

packages/tiled/lib/src/tile_map_parser.dart
```diff
@@ -10,12 +10,19 @@ class TileMapParser {
   ///
   /// Accepts an optional list of external TsxProviders for external tilesets
   /// referenced in the map file.
-  static TiledMap parseTmx(String xml, {List<TsxProvider>? tsxList}) {
+  static TiledMap parseTmx(
+    String xml, {
+    List<TsxProvider>? tsxList,
+    Future<TsxProvider> Function(String key)? tsxProviderFunction,
```
luanpotter, line 16: what's the difference between these two? seems like two different ways to pass in external tsx files?

spydon:
@trunghq3101 any update on this?

trunghq3101:
> @trunghq3101 any update on this?

I'm gonna continue working on this. I'm kinda being distracted by working on my game lately.

trunghq3101:
> > But according to the current Tiled format, template is not a child node anymore
> 
> Can you clarify what docs and when this changed? If it changed relatively recently we may want to check for the template xml the old way, and then if it's not there, check the new way. That would support both versions of Tiled. I would also add inline comments that link to the relevant Tiled docs and version changes

According to [this change log](https://github.com/mapeditor/tiled/blob/master/docs/reference/tmx-changelog.rst#tiled-11), `template` has been just a field of an `object` since day one. I didn't take a look at this when creating this PR. Is the any chance that our parser has never actually worked?

spydon:
> According to [this change log](https://github.com/mapeditor/tiled/blob/master/docs/reference/tmx-changelog.rst#tiled-11), `template` has been just a field of an `object` since day one. I didn't take a look at this when creating this PR. Is the any chance that our parser has never actually worked?

Sorry for the extremely late reply, it should be pretty simple to verify that with a small test, right?

spydon:
@trunghq3101 are you planning to continue on this PR? :)

trunghq3101:
> you

I guess I'm unable to do it anytime soon. Thanks for pinging me. Closing it.

PR #66, opened 2023-03-27
feat: Add convenience method for getting images in each layer

Hwan-seok:
# Description

This PR adds the convenience method to the TiledMap that gets all TiledImage in the specific layer.

If the layer is Group, the method gets all images in the group recursively.

## Checklist

## Breaking Change

## Related Issues

N/A

### DIFF
--- a/packages/tiled/lib/src/tiled_map.dart
+++ b/packages/tiled/lib/src/tiled_map.dart
@@ -219,6 +219,26 @@ class TiledMap {
     return imageSet.toList();
   }
 
+  List<TiledImage> getImagesInLayer(TileLayer layer) {
+    final usedTilesets = <Tileset>{};
+    final inspectedGids = <int>{};
+    const emptyTile = 0;
+    layer.tileData?.forEach((ty) {
+      ty.forEach((gid) {
+        if (gid.tile == emptyTile || inspectedGids.contains(gid.tile)) {
+          return;
+        }
+        inspectedGids.add(gid.tile);
+        usedTilesets.add(tilesetByTileGId(gid.tile));
+      });
+    });
+
+    return usedTilesets
+        .map((e) => [e.image, ...e.tiles.map((e) => e.image)].whereNotNull())
+        .expand((images) => images)
+        .toList();
+  }
+
   /// Finds the first layer with the matching [name], or throw an
   /// [ArgumentError] if one cannot be found.
   /// Will search recursively through [Group] children.
--- a/packages/tiled/lib/src/tiled_map.dart
+++ b/packages/tiled/lib/src/tiled_map.dart
@@ -219,24 +219,32 @@ class TiledMap {
     return imageSet.toList();
   }
 
-  List<TiledImage> getImagesInLayer(TileLayer layer) {
-    final usedTilesets = <Tileset>{};
-    final inspectedGids = <int>{};
-    const emptyTile = 0;
-    layer.tileData?.forEach((ty) {
-      ty.forEach((gid) {
-        if (gid.tile == emptyTile || inspectedGids.contains(gid.tile)) {
-          return;
-        }
-        inspectedGids.add(gid.tile);
-        usedTilesets.add(tilesetByTileGId(gid.tile));
+  List<TiledImage> collectImagesInLayer(Layer layer) {
+    if (layer is ImageLayer) {
+      return [layer.image];
+    } else if (layer is Group) {
+      return layer.layers.map(collectImagesInLayer).expand((e) => e).toList();
+    } else if (layer is TileLayer) {
+      final usedTilesets = <Tileset>{};
+      final inspectedGids = <int>{};
+      const emptyTile = 0;
+      layer.tileData?.forEach((ty) {
+        ty.forEach((gid) {
+          if (gid.tile == emptyTile || inspectedGids.contains(gid.tile)) {
+            return;
+          }
+          inspectedGids.add(gid.tile);
+          usedTilesets.add(tilesetByTileGId(gid.tile));
+        });
       });
-    });
 
-    return usedTilesets
-        .map((e) => [e.image, ...e.tiles.map((e) => e.image)].whereNotNull())
-        .expand((images) => images)
-        .toList();
+      return usedTilesets
+          .map((e) => [e.image, ...e.tiles.map((e) => e.image)].whereNotNull())
+          .expand((images) => images)
+          .toList();
+    }
+
+    return [];
   }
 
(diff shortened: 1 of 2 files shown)


Hwan-seok:
As of https://github.com/flutter/flutter/issues/91107, the tests are failing(technically, they could not be loaded).

Hwan-seok:
The test code action uses `flutter test` by the
https://github.com/flame-engine/flame-test-action/blob/fdf4cd9cf047b1b5e1359cf89ad0d88bca535523/scripts/test.sh#L10

Hwan-seok:
OK, I replaced the `test` to `flutter_test`.

packages/tiled/lib/src/tiled_map.dart
```diff
@@ -219,6 +219,34 @@ class TiledMap {
...
+    if (layer is ImageLayer) {
+      return [layer.image];
+    } else if (layer is Group) {
+      return layer.layers.map(collectImagesInLayer).expand((e) => e).toList();
+    } else if (layer is TileLayer) {
+      final usedTilesets = <Tileset>{};
+      final inspectedGids = <int>{};
+      const emptyTile = 0;
+      layer.tileData?.forEach((ty) {
+        ty.forEach((gid) {
```
jtmcdole, line 232: for (var gid in ty) {
  if (invalid/empty) continue;
}

packages/tiled/lib/src/tiled_map.dart
```diff
@@ -219,6 +219,34 @@ class TiledMap {
     return imageSet.toList();
   }
 
+  List<TiledImage> collectImagesInLayer(Layer layer) {
+    if (layer is ImageLayer) {
+      return [layer.image];
+    } else if (layer is Group) {
+      return layer.layers.map(collectImagesInLayer).expand((e) => e).toList();
```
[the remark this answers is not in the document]
Hwan-seok: Would you kindly explain the reason why you prefer the former one? :)
```dart
return [
  for (var layer in layer.layers.map(collectImagesInLayer)) ...layer
];

return layer.layers.map(collectImagesInLayer).expand((e) => e).toList();
```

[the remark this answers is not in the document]
spydon: You should also be able to just use `flattened` from the collection library: https://pub.dev/documentation/collection/latest/collection/IterableIterableExtension/flattened.html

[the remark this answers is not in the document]
Hwan-seok: > flattened

Very interesting one, I applied it.

packages/tiled/lib/src/tiled_map.dart
```diff
@@ -219,6 +219,36 @@ class TiledMap {
     return imageSet.toList();
   }
 
+  List<TiledImage> collectImagesInLayer(Layer layer) {
+    if (layer is ImageLayer) {
+      return [layer.image];
+    } else if (layer is Group) {
+      return layer.layers.map(collectImagesInLayer).flattened.toList();
```
luanpotter, line 226: nit: I think you can use `expand` instead of `map + flattened`

packages/tiled/lib/src/tiled_map.dart
```diff
@@ -219,6 +219,36 @@ class TiledMap {
...
+
+          if (tileGid == emptyTile || inspectedGids.contains(tileGid)) {
+            continue;
+          }
+          inspectedGids.add(tileGid);
+          usedTilesets.add(tilesetByTileGId(tileGid));
+        }
+      }
+
+      return usedTilesets
```
luanpotter, line 243: I think this could be

```dart
  final rows = layer.tileData ?? <List<Gid>>[];
  final gids = rows.expand((row) => row.map((gid) => gid.tile)).where((gid) => gid != emptyTile).toSet();
  return gids.map(tilesetByTileGId) ...
```

packages/tiled/lib/src/tiled_map.dart
spydon: Haha I love how everyone had their own suggestion, this seems like the best one though. :)

packages/tiled/lib/src/tiled_map.dart
Hwan-seok: I am trying to decide whether or not to divide the cascading in the middle.
Would you kindly give some thoughts on this?
```dart
      return rows
          .expand((row) => row.map((gid) => gid.tile))
          .where((gid) => gid != emptyTile)
          .toSet()
          .map(tilesetByTileGId)
          .toSet()
          .expand(
            (tileset) =>
                [tileset.image, ...tileset.tiles.map((tile) => tile.image)],
          )
          .whereNotNull()
          .toList();
```

```dart
      final gids = rows
          .expand((row) => row.map((gid) => gid.tile))
          .where((gid) => gid != emptyTile)
          .toSet();

      return gids
          .map(tilesetByTileGId)
          .toSet()
          .expand(
            (tileset) =>
                [tileset.image, ...tileset.tiles.map((tile) => tile.image)],
          )
          .whereNotNull()
          .toList();
```

packages/tiled/lib/src/tiled_map.dart
Hwan-seok: I realized that I had always been using `map + expand` rather foolishly 😂

spydon (approved):

packages/tiled/lib/src/tiled_map.dart
spydon: The second one might give slightly more readability.
You shouldn't have to do `toSet` again after `map(tilesetByTileGId)`, since the gid list already is unique.

Hwan-seok: > You shouldn't have to do toSet again after map(tilesetByTileGId), since the gid list already is unique.

The gid list is unique but they can be nonunique after `map(tilesetByTileGId)` because different gid can have the same tileset, doesn't it?

spydon: Ah right, true :)

packages/tiled/lib/src/tiled_map.dart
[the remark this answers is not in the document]
jtmcdole: I'm a fan of `yield` and `yield*`, but it's one of those "never really used" features. IIRC; its a sticking point for some WASM optimizations (please don't quote me).  Calling `map` makes a function call for each one, followed by expand(tear away that doesn't nothing but echos), and then a toList()... vs the Dart spread-operator with a map call.  I find the inline for(...) just easier to read.

Inline for

PR #72, opened 2023-10-30
feat!: Resolve external tilesets and templates through `ParserProvider` and add async `TiledMap.fromString`

### LINKED ISSUE #70: Test Function on custom TsxProvider`s
# What could be improved
Instead of checking by filename TsxProvider (or a new base class) could provide a function to check if a source can be resolved by the Provider.

# Why should this be improved
As I have mentioned in #69 (nice) I am currently working on a CLI to optimize tmx files. For this I resolve tsx files relative to the respective tmx file. Currently this has to be done by parsing to XML beforehand and resolving the tsx locations and then generating a TsxProvider for each. This seems rather inefficient

# Any risks?
I believe that this carries minmal risk, since the current functionality could be easily kept as a implementation of a new TsxProviderBase. Therefore no public contracts, naming or functionality would need to change, while greatly extending the capability of custom TsxProvider`s. 

# More information
I would gladly prepare a PR for this and improving documentation both in the README.md (which currently is wrong) and in code to explain this behaviour! I have already implemented this in the "#70-tsx-provider" branch of my [fork](https://github.com/benni-tec/tiled_dart/tree/%2370-tsx-provider).

### LINKED ISSUE #73: Template parsing does not work
# Current bug behaviour
The `TiledObject.parse()` function tries to find a `<template>` child of `<object>`. However according to the TMX Documentation and the dartdoc comment above the class such an element is not allowed as a child.

# Expected behaviour
Tiled uses the template reference in the `template` attribute of `<object>` to find the external template file. Maybe a lookup mechanism like TsxProviders is needed. 

# More information
Together with the extension to TsxProviders in #70 it might be worth generalizing a Provider for Tsx, template files and images.
I would be interested in working on a PR for this, however I do not need this functionality atm, so I wouldn't get to it for a few weeks.

benni-tec:
# Description

Replaces `TsxProvider` with a general `ParserProvider` that resolves any external file referenced from a map, and adds an async entry point for loading those files.

## Providers

```dart
abstract class ParserProvider {
  bool canProvide(String path);
  Parser getSource(String path);
}
```

- A single `providers` list is passed to `TiledMap.parseTmx` and `TiledMap.parseJson`. For every reference to an external file the first provider that `canProvide` the path is asked for a `Parser` through `getSource`. The path is always relative to the map (paths written in external files are joined with that file's directory first), so one provider can resolve any number of files (for example everything below a directory).
- Every external file is requested and parsed at most once per map, and cyclic references are left unresolved instead of recursing.
- The providers are carried on the `Parser`, so `Tileset.parse` and `TiledObject.parse` resolve external tilesets and object templates themselves, also inside already resolved files (template → tileset, tsx tile collision object → template).
- `TiledObject` gets a `templatePath` and its `template` is now parsed from the referenced template file instead of a non-existent `template` child element.
- `Parser.fromString` creates an `XmlParser` or a `JsonParser` depending on the contents.

## Async loading

```dart
final map = await TiledMap.fromString(
  contents,
  (path) => bundle.loadString('assets/tiles/$path'),
);
```

`TiledMap.fromString` parses either tmx or json and loads every referenced external file, recursively, with the given loader before returning the map. Parsing itself stays synchronous. Every file is loaded at most once.

## Other changes

- JSON maps now parse their tilesets (`tilesets` key and the string `image` with `imagewidth`/`imageheight`), so providers work for json maps too. Previously json maps had zero tilesets. Image layers get the same treatment, `parseJson` used to crash on them.
- `TileMapParser.parseTmx`/`parseJson` moved to `TiledMap.parseTmx`/`parseJson`.
- Objects inside template files do not require an `id`, which they never have.

## Checklist

## Breaking Change

### Migration instructions

- `TileMapParser.parseTmx(xml, tsxList: [...])` → `TiledMap.parseTmx(xml, providers: [...])`, `TileMapParser.parseJson(json)` → `TiledMap.parseJson(json)`.
- `TsxProvider` → `ParserProvider`. Replace the `filename` getter with `canProvide(path)`, drop `getCachedSource` (cache inside `getSource` if desired) and return a `Parser` from `getSource(path)`, for example with `Parser.fromString`.
- `TiledMap.fromString(contents, tsxProviderFunction)` → `TiledMap.fromString(contents, loadFile)` where `loadFile` returns the file contents as a `Future<String>`. `flame_tiled` can drop `FlameTsxProvider` and pass `(path) => bundle.loadString('$prefix$path')`.
- `Tileset.parse(parser, tsx: ...)` → `Tileset.parse(parser)`, external tilesets are resolved through the parser's providers.

## Related Issues

Closes #70
Closes #73


benni-tec:
I've got some questions about the checklist:
- The old functionality was not covered by tests as far as I can see, nevertheless I will write some soon.
- There is no `example` folder in this repo. Where should examples go?

kurtome:
>While do not like attaching Base to everything this naming scheme seems to be the easiest option to not break current implementations.

I would consider just breaking current implementations, as long as you are willing to write a PR for `flame_tiled`, which is the only major client of this library. 

Overall it's hard to me to tell if this PR makes sense without concrete examples, maybe tests would help with that

ufrshubham (approved):
I can't seem to understand why `TsxProviderBase` had to be a separate class. If it is just to avoid breaking `filename` getter, then I think making that breaking change should be fine.

README.md
```diff
@@ -29,9 +29,15 @@ Load a TMX file into a string by any means, and then pass the string to TileMapP
     final TiledMap mapTmx = TileMapParser.parseTmx(tmxBody);
 ```
 
-If your tmx file includes a external tsx reference, you have to add a CustomParser
+If your tmx file includes a external tsx reference, you have to add a CustomParser. This can either be done by using extending the TsxProviderBase, which can match multiple files, or by extending TsxProvider, which only matches on file by its name.
```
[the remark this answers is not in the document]
benni-tec: Whoops missed that on 😅

packages/tiled/lib/src/tsx_provider.dart
```diff
@@ -1,18 +1,44 @@
 part of tiled;
 
 /// abstract class to be implemented for an external tileset data provider.
-abstract class TsxProvider {
+abstract class TsxProviderBase {
+  /// Given a filename this function should check whether this Provider
+  /// can provide this source.
+  bool checkProvidable(String filename);
```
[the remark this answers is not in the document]
benni-tec: Yes that does sound better 👍

spydon:
@benni-tec any updates on this? :)

benni-tec:
Not yet. My fork currently does everything I need for my current project. I do plan to submit proper PRs for this and the other issues I've opened, however I am currently pretty time constrained. 

As of now my current project should be wrapped up in January, I am currently planning to clean up and submit PRs then.

Is this something that is time-sensitive for you? If so I will see if can get the PRs done this year!

packages/tiled/lib/src/template.dart
```diff
@@ -34,3 +34,7 @@ class Template {
           object: parser.getSingleChildOrNullAs('object', TiledObject.parse),
         );
 }
+
+class TemplateReference {
```
spydon, line 38: Should this be removed?

packages/tiled/lib/src/tiled_map.dart
```diff
@@ -125,28 +125,31 @@ class TiledMap {
...
+  //     tsxSourcePaths
+  //         .where((key) => key != null)
+  //         .map((key) async => tsxProviderFunction(key!)),
+  //   );
+  //
+  //   return TileMapParser.parseTmx(
+  //     contents,
+  //     tsxList: tsxProviders.isEmpty ? null : tsxProviders,
+  //   );
+  // }
```
spydon, line 152: Remove?

packages/tiled/lib/src/tiled_map.dart
```diff
@@ -286,7 +289,41 @@ class TiledMap {
     );
   }
 
-  factory TiledMap.parse(Parser parser, {List<TsxProvider>? tsxList}) {
+  static TiledMap parseJson(
```
spydon, line 292: This should be a factory just like `parseTmx`, also needs some dartdocs

spydon (commented):
Just a few comments, over all it looks good :)

jtmcdole:
LFTM (looks fine to me!)

benni-tec:
Thanks for the review, I didn't have to time yet to write a proper comment: 

I have generalised into a generic Provider that can also be used for templates, therby also resolving #73, also as being extensible to more easily discover the location of images in `flame_tiled`.

I also moved the methods from TileMapParser into TiledMap since there were already parsing method there, and I always thought it was confusing to have this in a different class that does nothing otherwise. While moving it over I also made the names nicer and more descriptive.

The `fromString` method seems to only have been used in the tests to generate `TsxProvider`'s for every included tileset, which always seemed like a hack to me! Anyways this is not neccessary anymore since the new Provider can resolve multiple or even all files (see tests for an example). The rest of the functionality can be found in `parseTmx()`!

I took the liberty of adding the Lists of `Provider`'s to the parser to make it more convenient and clean (instead of passing it through all parse calls, only the initial Parser creation is affected and the copies)! 

However I split them into multiple Lists so one for tmx and one for templates. This however does not allow to integrate the same mechanism for images (e.g. of tilesets), because I don't think the actual loading of the images is supposed to be in this library and nothing is ever done with them I don't think that is a big deal. `flame_tiled` could still use the same `Provider` mechanism to discorver images (this would be great improvment compared to the current state), but just create it's own List when parsing the RenderableMap!

However I could also only pass one List and the `Provider`'s could instead report their capability. What do you think is preferable?

spydon:
Sorry for the extremely late reply @benni-tec, I had completely forgotten that we had open PRs on this repository!
What is the state of your PRs, are they ready to be merged?

benni-tec:
Honestly I also forget about the PRs since my project was finished in January, I just merged the incoming changes into #77 and it should be ready to merge now.

I will merge theses changes here as well and incorporate the feedback today, this PR should then also make #59 obsolete.

As for #78 I believe you said it should be an extension package, however I beliebe I made some adjustments in that PR that should be merged to, I will have to look at it again

packages/tiled/lib/src/template.dart
benni-tec: Probably 😂

benni-tec:
Alright so I rebased this branch onto #77 and fixed the docs as well as incorporating the feedback here. The tests also clear now, as for the analyze it failed due to the partof directives so I'm gussing thats fine? And I don't know whats up with format

spydon:
@benni-tec could you do a final check on the PR just to see that everything looks correct after the rebase? :)

spydon:
Ping @benni-tec, I think this is pretty much ready to merge right?

benni-tec:
Hi everybody, I kept this discussion in my inbox (and the back of my mind) but never got around to checking-in again, sorry!

I just had a look and I believe it is indeed ready to be merged, the only thing that is missing is the appropriate PR for `flame_tiled`. Those changes should be minimal though.

If you are still interested in this I will try to get it done before the new year.

P.S.: I will be closing the export PR and try to move the code to a new repository. That is indeed a lot of code and potentially a huge maintanence burden

benni-tec:
While integrating this into Flame I discovered a new problem: Loading from the bundle is asynchronous.

Therefore either the provider and thus the parsing needs to be asynchronous or we hack it in `flame_tiled` again by crawling the xml and loading all (?) of the files from the bundle. I believe this is how it was before therefore performance should not get worse. Alternativley we could pass a list of what files should be loaded.

If the provider and therefore the parser should stay synchronouse (which is a valid point IMO), then the `getCachedSource()` only makes sense for synchronous I/O. This is bad practice anyway so we could probably just remove it.

spydon:
@benni-tec sounds good, go for it. :)

benni-tec:
What would you prefer:
- Asynchronous Parsing
- Hack in flame_tiled which only supports xml (as it is now)

spydon:
@benni-tec sorry for the extremely late reply, I missed this notification. Async parsing should be fine.

ufrshubham (approved):
LGTM!

PR #74, opened 2023-11-02
fix: `ObjectAlignment` enum names

ufrshubham:
# Description

Name of some of the enum values of `ObjectAlignment` were not matching the values from Tiled. As a result, tilesets with `topLeft`, `topRight`, `bottomLeft` and `bottomRight` alignment couldn't be parsed. https://doc.mapeditor.org/en/stable/reference/tmx-map-format/#tmx-tileset

image

Another problem that was discovered after fixing the names was with the `byName` method from the `EnumByName` extension. That method was not using names from `ObjectAlignmentExtension` that Tiled defines. To fix that, this PR adds a `byName` static method for `ObjectAlignment` which uses the correct names.

## Checklist

## Breaking Change

## Related Issues

### DIFF
--- a/packages/tiled/lib/src/common/enums.dart
+++ b/packages/tiled/lib/src/common/enums.dart
@@ -386,7 +386,7 @@ enum ObjectAlignment {
   /// Returns the [ObjectAlignment] based on given [name].
   ///
   /// Throws an [ArgumentError] if no match is found.
-  static ObjectAlignment byName(String name) {
+  static ObjectAlignment fromName(String name) {
     for (final value in ObjectAlignment.values) {
       if (value.name == name) {
         return value;
--- a/packages/tiled/lib/src/common/enums.dart
+++ b/packages/tiled/lib/src/common/enums.dart
@@ -373,15 +373,15 @@ extension TilesetTypeParser on Parser {
 
 enum ObjectAlignment {
   unspecified,
-  topleft,
+  topLeft,
   top,
-  topright,
+  topRight,
   left,
   center,
   right,
-  bottomleft,
+  bottomLeft,
   bottom,
-  bottomright,
+  bottomRight,
 }
 
 extension ObjectAlignmentExtension on ObjectAlignment {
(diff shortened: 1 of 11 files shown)


jtmcdole:
Would you want to leave the enum symbols in `lowerUpperCase` to pass lint tests, but return the string `lowercase`?

ufrshubham:
> Would you want to leave the enum symbols in `lowerUpperCase` to pass lint tests, but return the string `lowercase`?

The lint warning were not for the enum change, but I made the change anyways 🙂

ufrshubham:
@spydon I noticed that the `pubspec.yaml` was missing from the root of this project. If I am not wrong, it needs to be added because of melos. I added it in this PR, but now a lot of warnings are reported. Should I make these additional changes in a separate PR or pushing it along with this PR is fine?

spydon:
@ufrshubham aah yes, that is needed for the "new" Melos version. You can fix them in this PR if it is easier for you. If you upgrade flame_lint to 1.1.1 the warning should go away.

packages/tiled/lib/src/common/enums.dart
```diff
@@ -381,7 +381,19 @@ enum ObjectAlignment {
   right,
   bottomLeft,
   bottom,
-  bottomRight,
+  bottomRight;
+
+  /// Returns the [ObjectAlignment] based on given [name].
+  ///
+  /// Throws an [ArgumentError] if no match is found.
+  static ObjectAlignment byName(String name) {
```
spydon, line 389:
```suggestion
  static ObjectAlignment fromName(String name) {
```
Maybe?

spydon (approved):

packages/tiled/lib/src/common/enums.dart
ufrshubham: Renamed to `fromName`

spydon (approved):

spydon:
Haha aah, unlucky with one line going past 80 now :sweat_smile:

PR #75, opened 2023-12-05
feat: Adding a method to get any object in a map by its unique ID

kurtome:
# Description

Adding a method to get any object in a map by its unique ID, `map.objectById`

## Checklist

## Breaking Change

### DIFF
--- a/packages/tiled/lib/src/tiled_map.dart
+++ b/packages/tiled/lib/src/tiled_map.dart
@@ -91,6 +91,9 @@ class TiledMap {
   List<EditorSetting> editorSettings;
   CustomProperties properties;
 
+  // Cache the object by ID when accessed.
+  Map<int, TiledObject>? _cachedObjects;
+
   // only for hexagonal maps:
   int? hexSideLength;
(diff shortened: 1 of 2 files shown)


packages/tiled/lib/src/tiled_map.dart
```diff
@@ -279,6 +282,24 @@ class TiledMap {
     throw ArgumentError('Layer $name not found');
   }
 
+  /// Finds the [TiledObject] in this map with the unique [id].
```
ufrshubham, line 285: Might as well mention that the first call will cache all the objects. Based on how complex the map is, users might want to avoid making the first call during the gameplay. It is highly unlikely, but we never know 😄

ufrshubham (approved):

packages/tiled/lib/src/tiled_map.dart
kurtome: 👍 good idea, done

kurtome:
@spydon I think this is ready to merge 👍 , I don't have access on this repo

Issue #69, opened 2023-10-28
Dart SDK compatibility


benni-tec:
# What could be improved
Remove requirement for `dart:ui` and therefore the Flutter SDK

# Why should this be improved
Currently this package/parser is only usable with the Flutter SDK. However I believe there are valid use-cases for the Dart SDK. For example I'm currently working on a CLI to optimize tmx to only include the tiles actually used in a map in it's tilesets. This is not possible using the Flutter SDK.
 
# Any risks?
This improvement would require the removal of all dart:ui dependencies. However there are only two in this project, which I replaced in the "#69-dart-sdk branch" on my [fork](https://github.com/benni-tec/tiled_dart/tree/%2369-dart-sdk):
- Color, which I replaced with [color_models](https://pub.dev/packages/color_models) and a little bit of code for hex parsing
- Rect, which I simply replaced by `dart:math`'s Rectangle

# More information
These changes would break public contracts and therefore require some work on the flame_tiled plugin. 

However I believe this would be minimal. I did not look into this further, but from what I changed this would only require converting Rect to Rectangle and casting ColorModel to Model, which is implemented in [flutter_color_models](https://pub.dev/packages/flutter_color_models).

I believe this would be worth it and I am prepared to submit PRs both here and in flame_tiled to make these rather simple changes. However if you would consider this to be too invasive, I am happy to just keep this as a fork!

spydon:
Overall it sounds like a good idea, we've had this plan before but never done the implementation.
I'm was thinking that maybe it is a better idea to pull in the color and rect class directly in tiled.dart instead of pulling it from the third party library so that we don't have to make yet another breaking change if that library become unmaintained an we have to update something, but on the other hand we could just pull in those classes at that time then instead.

spydon:
@kurtome @ufrshubham @jtmcdole do you have any input on this, #70 and #71?

benni-tec:
> Overall it sounds like a good idea, we've had this plan before but never done the implementation. I'm was thinking that maybe it is a better idea to pull in the color and rect class directly in tiled.dart instead of pulling it from the third party library so that we don't have to make yet another breaking change if that library become unmaintained and we have to update something, but on the other hand we could just pull in those classes at that time then instead.

Maybe I'm missing something, but nothing is ever done with both Rect and Color right? 
Could Color than not just store RGBA values and thats it? There would be no need for any library at all.
(Don't know why I did not think of this before when I implemented this)

[...]

PR #77, opened 2024-01-16
feat!: Dart SDK compatibility 

benni-tec:
# Description
This PR makes this package compatible with the Dart SDK. In order to achieve this I changed 3 things:
- Replace `dart:ui`'s Color with a simple ColorData class
- Replaced `Rect` from `flame` with `Rectangle` from `dart:math`
- Replaced `flutter_test` with `test` package

The `ColorData` class is used excatly once for which I prepared a flame_tiled PR (however I need to wait until this is merged right) including conversion extensions.

No `Rectangle` attribute is ever read in `flame_tiled`, evenso `flame` already contains conversion extensions!

## Checklist
  --> `melos run analyze` fails due to `XmlData.value` however I think this requires another issue/PR
  --> I don't think there are any exmaples necessary for this

## Breaking Change

I think this should result in a minor version bump to 0.11.0 to indicate these breaking changes!

In order to achieve dart comaptibilty only two things changed:
- Color: Custom data class --> `flame_tiled` contains a conversion extension
- Rect: Instead using `dart:math`'s Rectangle --> Use `flame`'s `toRect()` method to convert

## Related Issues
- #69


spydon (approved):
Lgtm!

packages/tiled/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: tiled
-version: 0.10.1
+version: 0.11.0
```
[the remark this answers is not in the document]
benni-tec: Fixed 👍

packages/tiled/lib/src/common/color.dart
```diff
@@ -0,0 +1,25 @@
+part of tiled;
+
+/// Basic data class holding a Color in ARGB format.
+/// This can be converted to dart:ui's Color using the flame_tiled package
+class ColorData {
+  static int _sub(int hex, int index) => (hex >> index * 8) & 0x000000ff;
```
jtmcdole, line 6: I just saw this in the other PR; is there a mix up between the two? If so; my comment there is valid here as well: why not store the "hex" as one value and do bit shifting in setters?

This is what `Color` was doing before and so the pattern would match and the memory footprint should be the same.

packages/tiled/lib/src/common/color.dart
```diff
@@ -0,0 +1,25 @@
...
+  final int alpha;
+
+  /// Parses the Color from an int using the lower 32-bits and tiled's format: 0xaarrggbb
+  ColorData.hex(int hex)
+      : alpha = _sub(hex, 3),
+        red = _sub(hex, 2),
+        green = _sub(hex, 1),
+        blue = _sub(hex, 0);
+
+  const ColorData.rgb(this.red, this.green, this.blue, [this.alpha = 255])
```
jtmcdole, line 20: "rgba"? and why deviate from "argb"?

packages/tiled/lib/src/common/color.dart
benni-tec: Yes you are right that would be more efficient! This change does belong here, but I based the PR‘s branch on this one so I don‘t have to integrate the new types later!

packages/tiled/lib/src/common/color.dart
benni-tec: I don’t really know, it just felt more natural. For most colors the alpha value is just 0xff so it can have a default value at the end. CSS also puts it at the end #rrggbbaa, I believe for the same reason. 

This constructor is meant for users and not used internally, so I think this way is a better experience. However I‘m happy to change it!

jtmcdole: Right, I was trying to stick with the dart:ui color for familiarity.

kevmoo:
Would this help here? https://flutter.dev/go/move-flutter-agnostic-types

If yes, help me understand how.
If not, what's missing?

spydon:
> Would this help here? https://flutter.dev/go/move-flutter-agnostic-types
> 
> If yes, help me understand how.
> If not, what's missing?

Yes! That would be exactly what we need here, it migrates all the types that we need out of flutter.

Thanks!

spydon:
@benni-tec is this ready to be merged?
Because we aren't getting these classes (Color, Rect etc) separated any time soon right @kevmoo?

benni-tec:
I believe it should be ready to merge now, I adjusted Color to fit @jtmcdole's suggestions.

spydon (approved):
I did some final clean-up so that the analyzer and tests pass (also on non-windows machines), now I think we can finally merge this! :partying_face:

kevmoo:
> @benni-tec is this ready to be merged? Because we aren't getting these classes (Color, Rect etc) separated any time soon right @kevmoo?

That work was on hold until the "big merge" completed. Hoping to jump in again in the new year!

